PR #1, opened 2013-11-16
Fix partially bug #3394 (https://cfengine.com/dev/issues/3394) ; use sta...

fabix:
...rt-stop-daemon with its pidfile argument.


jimis:
@kacfengine maybe you can have a look in this one, since you recently reviewed our initscripts?

kacfengine:
Right, this was forgotten, but it looks alright. I have resubmitted it here: https://github.com/cfengine/core/pull/2139.

PR #34, opened 2015-01-07
Redmine #6818: Detect and clean up lingering processes in chroot.

kacfengine:
Note that we consider failure to clean up after yourself an error.
The chroot survives it just fine, but we still want the tests to
behave nicely so that it doesn't happen when you run them locally
either.


kacfengine:
trigger build

PR #38, opened 2015-01-21
Lower down the limit for using optimised postgresql.conf from 8GB to 3GB...

maciejmrowiec:
... of total RAM.

Redmine: #6936


maciejmrowiec:
@kacfengine could you have a look? this is 3.6.4 stuff.

PR #40, opened 2015-01-23
build-remote now passes the CFLAGS and CC env vars to the remote buildslave

jimis:


jimis:
Pretty trivial, so I'm merging and cherry-picking!

kacfengine:
> Untabify and reindent with 2 spaces to be consistent.

+1! We have many scripts with such inconsistencies and they always bug me.

build-scripts/functions
```diff
@@ -226,6 +226,18 @@ remote_script_general() {
...
+  fi
+  if [ -n "$CPPFLAGS" ]; then
+    ENVVARS="$ENVVARS CPPFLAGS='$CPPFLAGS'"
+  fi
+  if [ -n "$CFLAGS" ]; then
+    ENVVARS="$ENVVARS CFLAGS='$CFLAGS'"
+  fi
+  if [ -n "$LDFLAGS" ]; then
+    ENVVARS="$ENVVARS LDFLAGS=$LDFLAGS"
+  fi
```
kacfengine: This should be documented on the help screen. If you happen to have them set in your environment you are going to get very strange errors, and your flags will almost certainly be wrong for the platform unless you intended it.

build-remote
```diff
@@ -46,12 +47,13 @@ opts() {
   BRANCH=master
   BUILD_TYPE=quick
   SOURCE=git@github.com:cfengine
-  BUILD_NUMBER=${BUILD_NUMBER:-666}
+  BUILD_NUMBER=${BUILD_NUMBER:-unknown}
```
kacfengine: +1!

build-scripts/functions
jimis: On Fri, Jan 23, 2015 at 1:32 PM, Kristian Amlie <notifications@github.com>
wrote:

> In build-scripts/functions
> <https://github.com/cfengine/buildscripts/pull/40#discussion_r23446925>:
>
> > @@ -226,6 +226,18 @@ remote_script_general() {
> >    if [ -n "$PREFIX" ]; then
> >      ENVVARS="$ENVVARS BUILDPREFIX=$PREFIX"
> >    fi
> > +  if [ -n "$CC" ]; then
> > +    ENVVARS="$ENVVARS CC=$CC"
> > +  fi
> > +  if [ -n "$CPPFLAGS" ]; then
> > +    ENVVARS="$ENVVARS CPPFLAGS='$CPPFLAGS'"
> > +  fi
> > +  if [ -n "$CFLAGS" ]; then
> > +    ENVVARS="$ENVVARS CFLAGS='$CFLAGS'"
> > +  fi
> > +  if [ -n "$LDFLAGS" ]; then
> > +    ENVVARS="$ENVVARS LDFLAGS=$LDFLAGS"
> > +  fi
>
> This should be documented on the help screen. If you happen to have them
> set in your environment you are going to get very strange errors, and your
> flags will almost certainly be wrong for the platform unless you intended
> it.
>
Will document and push directly.

jimis:
> >    Untabify and reindent with 2 spaces to be consistent.

> +1! We have many scripts with such inconsistencies and they always bug me.

Yeah but this change was only a drop in the ocean. I would suggest we apply our C coding style to all our scripts, i.e. 4 spaces, no tabs, begin/end on separate lines.

kacfengine:
> I would suggest we apply our C coding style to all our scripts, i.e. 4 spaces, no tabs, begin/end on separate lines.

I'm fine with that. I've been using two spaces for a long time because it seemed like the standard to me (from the scripts I had seen), but in reality it's a mess, and I'm all for making it consistent with the C code.

PR #41, opened 2015-01-23
Document env variables transferred to the buildslave

jimis:


build-scripts/functions
```diff
@@ -244,7 +244,7 @@ remote_script_general() {
   ENVVARS="$ENVVARS EXPLICIT_VERSION=$EXPLICIT_VERSION"
   ENVVARS="$ENVVARS TEST_MACHINE=$TEST_MACHINE"
 
-  $LOGIN_COMMAND env $ENVVARS "$SCRIPT_BASEDIR"/buildscripts/build-scripts/"$SCRIPT"
+  ( set -x; $LOGIN_COMMAND env $ENVVARS "$SCRIPT_BASEDIR"/buildscripts/build-scripts/"$SCRIPT" )
```
kacfengine: Are you sure this works? Is it really inherited by SSH?

build-remote
```diff
@@ -26,6 +26,9 @@ usage() {
   echo "Build type defaults to quick. You may request a full build using --full"
   echo "Source defaults to git@github.com:cfengine/"
   echo
+  echo "NOTE: The following environment variables will be carried to the remote machine:"
+  echo "      CC CPPFLAGS CFLAGS LDFLAGS"
+  echo
```
kacfengine: +1!

build-scripts/functions
jimis: On Fri, Jan 23, 2015 at 2:40 PM, Kristian Amlie <notifications@github.com>
wrote:

> In build-scripts/functions
> <https://github.com/cfengine/buildscripts/pull/41#discussion_r23449592>:
>
> > @@ -244,7 +244,7 @@ remote_script_general() {
> >    ENVVARS="$ENVVARS EXPLICIT_VERSION=$EXPLICIT_VERSION"
> >    ENVVARS="$ENVVARS TEST_MACHINE=$TEST_MACHINE"
> >
> > -  $LOGIN_COMMAND env $ENVVARS "$SCRIPT_BASEDIR"/buildscripts/build-scripts/"$SCRIPT"
> > +  ( set -x; $LOGIN_COMMAND env $ENVVARS "$SCRIPT_BASEDIR"/buildscripts/build-scripts/"$SCRIPT" )
>
> Are you sure this works? Is it really inherited by SSH?
>
To tell the truth I only ran a couple of tests on my local shell. It should
work (TM).

kacfengine: Actually this is more likely to be the problem of '-x' missing in some other file, since it should already be active here. IOW, I don't think this is the right fix. Which script was being called that you didn't see?

jimis: @kacfengine this will print it no matter the verbosity level. I think it's
a good see to always see that one line in the output.

kacfengine: Oh, you mean as a replacement/addition to the '--> <script-name>' thing? Yeah, ok, that makes sense.

kacfengine:
trigger build

kacfengine:
Triggering a build, since that function is very central.

PR #46, opened 2015-02-12
Fix cfengine-nova-hub package using scripts from cfengine-nova.

kacfengine:
Reflect that the fix in 9dbcdfebd9 really only is for cfengine-nova.


kacfengine:
Packages are broken in master, so merging directly.

PR #47, opened 2015-02-13
Fix Solaris binaries not finding the libnet library.

kacfengine:
The LD_LIBRARY_PATH points to Java libraries, and for some reason the
OpenLDAP build thinks it needs the libnet library from there, which is
of course completely wrong, since it is a Java library. Nevertheless
it tries to depend on it, which then causes problems when trying to
run the binary.


kacfengine:
trigger build

kacfengine:
trigger build

PR #52, opened 2015-03-05
Fix incorrect flag passed to linker, should have "-Wl," prefix.

kacfengine:
This fixes RHEL7, and I can't quite say how I got it to work before.


kacfengine:
trigger build

PR #56, opened 2015-04-04
Check PATH and MANPATH content before expanding them.

lukaf:
Check for existence of /var/cfengine/bin in $PATH and /var/cfengine/share/man in $MANPATH before expanding them.

My current $MANPATH, running inside tmux inside X terminal:
```bash
$ echo $MANPATH
:/var/cfengine/share/man:/var/cfengine/share/man:/var/cfengine/share/man
```


kacf:
Sorry, this was forgotten in the queue, I have merged it now. Thanks!

PR #59, opened 2015-04-16
Fix fuser output.

jimis:
fuser on RHEL4 is being printed differently. So avoid any parsing of its
output and just call fuser in verbose mode, which prints the process
names.

### DIFF
--- a/build-scripts/test-on-testmachine
+++ b/build-scripts/test-on-testmachine
@@ -35,24 +35,15 @@ fi
 
 case "$TEST_MACHINE" in
     chroot)
-        # Fuser has special output. The PIDs arrive on stdout, and all the ornaments
-        # arrive on stderr, so all we have to do is to grep for PID numbers.
-        for pid in `sudo $FUSER $CHROOT_ROOT 2>/dev/null`; do
-            # Leaving processes behind is an error. It should never happen.
-            return_code=1
-            (
-                # This next part is very confusing when mixed with "set -x".
-                # Turn it off inside the subshell while printing this.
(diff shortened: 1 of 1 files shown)


jimis:
trigger build

jimis:
Hmmm `fuser -v` does not show the arguments of processes, only the process names. Maybe I should add a `ps auxfww | grep -C2 cf-` in case `fuser` detects stale processes.

jimis:
did that

build-scripts/test-on-testmachine
```diff
@@ -35,24 +35,18 @@ fi
...
-                set +x
-                echo "Error: Found processes left behind in the chroot. I'm killing them, but the build will fail, so clean up your mess!"
-                echo "Found PID $pid. Info from 'ps -ef'"
-                echo "--------------"
-                ps -ef | grep $pid
-                echo "--------------"
-            )
-        done
+        echo "Checking for stale processes in the chroot path: $CHROOT_ROOT"
+        sudo $FUSER -v $CHROOT_ROOT 2>&1
```
kacfengine: ``-v`` is not universally supported. Besides we'd really like to get PIDs, because the process is not necessarily a cf-* process, for example the build that triggered this tested if we could start ntpd, which then got left running. I will make a separate fix.

jimis:
Is this stale-processes-detection running on non-linux platforms as well?

kacfengine:
Yes, it runs on all Unix platforms, otherwise we might leave processes behind for the next test run.

PR #60, opened 2015-04-17
Fix fuser output being misinterpreted on RHEL4.

kacfengine:
Most platforms print the PIDs on stdout, and decorative characters on
stderr. Together they look like nicely formatted output when printed
on the console, but it also allows you to pipe only the PIDs somewhere
else. However, RHEL4 prints everything on stdout, so we need to filter
out the superflous characters there. All we need are numbers and the
spaces between them, so reject everything else.


kacfengine:
trigger build

PR #66, opened 2015-05-28
Redmine #7161: Include package modules in packages.

kacfengine:


kacfengine:
Depends on https://github.com/cfengine/masterfiles/pull/465.

PR #71, opened 2015-06-12
Redmine#7270: iconv is needed to be built into PHP

jimis:
It's used (but probably shouldn't be) in a couple of places in
cf_util_helper.php.


jimis:
trigger build

kacfengine:
AFAIK, iconv is installed by default on Linux. Attempting with #72 instead.

PR #74, opened 2015-06-17
Redmine #7289: Add libltdl7 as a dependency to the hub package.

kacfengine:
All Debian platforms which support the hub package (Debian 6 and
higher, as well as Ubuntu 10 and higher) need this library in order
to work. Debian 4 doesn't have the library, but we also don't build a
hub package for that platform.


kacfengine:
trigger build

PR #75, opened 2015-06-19
Fix SSIZE_MAX not being defined in certain MinGW versions.

kacfengine:
The build slaves don't seem to need it, but I do on my machine.


kacfengine:
trigger build

PR #78, opened 2015-06-29
Bump postgres config to make autovacuum more aggressive.

maciejmrowiec:


kacfengine:
Cherry-picked to both 3.7.x and 3.6.x.

PR #80, opened 2015-07-08
Redmine #6954: Prevent accumulation of unrelated AIX packages.

kacfengine:
The directory is not cleaned, so each time we build a different
version we get a new package. This propogates all the way to the final
package directory, which is both confusing and potentially dangerous,
since someone may be tempted to download a package for a given version
that looks legitimate, but is in fact, an old/broken build.

### DIFF
--- a/packaging/cfengine-nova/cfengine-nova.bff.sh
+++ b/packaging/cfengine-nova/cfengine-nova.bff.sh
@@ -15,8 +15,14 @@ BASEDIR=$2
 LPPBASE=$2/..
 P="$BASEDIR/buildscripts/packaging/cfengine-nova"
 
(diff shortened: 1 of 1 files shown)


kacfengine:
trigger build

jimis:
+1, lets see how the triggered build goes.

kacfengine:
It didn't work, I still see numerous unrelated packages in the upload, I didn't expect that..

kacfengine:
trigger build

kacfengine:
Yey, problem solved!

kacfengine:
Cherry-picked to both 3.6.x and 3.7.x.

jimis:
Hurray!

PR #82, opened 2015-08-06
Add build-scripts/bootstrap-tarballs and build-scripts/unpack-tarballs

jimis:
The first is supposed to run from the "bootstrap" jenkins job and create
the core and masterfiles tarballs, and delete the git-checked-out
directories.

The second is supposed to run in the next stage in each of the
buildslaves, and unpack the tarballs and symlink into the core/ and
masterfiles/ directories.

This way we have a separate tarball from the "bootstrap" slave suitable
for releasing, and we test in all platforms that the tarball is valid.

### DIFF
--- a/build-scripts/bootstrap-tarballs
+++ b/build-scripts/bootstrap-tarballs
@@ -1,0 +1,30 @@
+#!/bin/sh -x
+
+. `dirname "$0"`/functions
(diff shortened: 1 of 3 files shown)


jimis:
It's safe until the jenkins jobs are modified, merging.

PR #87, opened 2015-09-24
Redmine #7591: Fix Solaris pkgs contain dupl. files instead of links.

kacfengine:
There were in fact three separate issues at work to prevent symlinks
from reaching the package:

The first is the fact that Solaris' shell appears to return true for
the -f test, even if the file is actually a link. So we need to test
for a link first. This is the fix in pkg-build-solaris.

Secondly, cp does not preserve links unless -P is given, but -P is not
supported on Solaris 9, so instead we use rsync with the -l option.

And third, we were using `find -type f` to iterate through the package
candidate files, but this would not return links at all. We switch to
using the pkgproto tool directly instead, but this required some
restructuring because the directory cannot have any non-package files
when that command is invoked, or they will be included in the package
verbatim. Changed the logic to copy these into the directory later
instead.

Changelog: Fix: Solaris packages no longer contain duplicate library
files, but instead symlinks to them.

### DIFF
--- a/build-scripts/package
+++ b/build-scripts/package
@@ -172,21 +172,27 @@ case "$PACKAGING" in
     sudo rm -rf $BASEDIR/$PKG/pkg
     mkdir -p $BASEDIR/$PKG/pkg
 
-    cp -pr $BASEDIR/cfengine/dist/* $BASEDIR/$PKG/pkg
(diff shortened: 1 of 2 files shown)


kacfengine:
trigger build

kacfengine:
trigger build

kacfengine:
Cherry-picked to 3.7.x as well.

PR #88, opened 2015-10-16
Redmine #6708: Fix RPATH on all platforms

kacfengine:
See commit log for individual fixes.

This has been tested together with the Enterprise unit test on a private branch, and passes all tests.


kacfengine:
Cherry-picked to 3.7.x as well.

PR #94, opened 2015-11-03
Old Ubuntus should only build agent package. Hub is no longer supported.

jimis:


kacfengine:
Should trigger build on this after the platform is taken out of Jenkins hub build.

jimis:
I have run the "manage_jenkins_job" bundle so it should be ready to trigger.

PR #95, opened 2015-11-04
Downgrade PostgreSQL to 9.3.9 again.

kacfengine:
This is a partial revert of 9862d044446f8ba. The new PostgreSQL
version has a different on-disk format, which requires cumbersome
migrations. It's not very important to upgrade right now, so we'll
stay on 9.3.9.


kacfengine:
trigger build

kacfengine:
Cherry-picked to 3.8.x as well.

PR #96, opened 2015-11-06
Revert "Redmine #6480: Enable libcurl for all platforms."

pasinskim:
This reverts commit eada84c335dbb52ca9c9ae52b10dc57ff879cdb6.

For checking if https login issues are related to curl upgrade.


kacfengine:
It just occurred to me that there is a test which tests curl functionality. I suppose it may fail if this is removed, and prevent the package from being built. Sorry, I didn't think about that earlier. Perhaps it's better to run it in a private branch and comment the "test" part of the build recipe?

pasinskim:
trigger build

pasinskim:
trigger build

pasinskim:
This is was for testing libcurl only and is not needed any more.

PR #97, opened 2015-11-13
Redmine #7267: Allow exporting CSV files over HTTPS.

pasinskim:
In order to support CSV reports being send over HTTPS connection
we need to have cert.pem certificate in /var/cfengine/httpd/ssl.

This one is making a symbolic link in mentioned location to the
certificate file placed in /var/cfengine/httpd/ssl/.

Changelog: Fix exporting CSV reports through HTTPS. (Redmine #7267)

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -233,15 +233,25 @@ done
 # The certificate will be named $(hostname -f).cert and the corresponding key should be named $(hostname -f).key.
 #
 CFENGINE_MP_DEFAULT_CERT_LOCATION="$PREFIX/httpd/ssl/certs"
+CFENGINE_MP_DEFAULT_CERT_LINK_LOCATION="$PREFIX/httpd/ssl"
 CFENGINE_MP_DEFAULT_KEY_LOCATION="$PREFIX/httpd/ssl/private"
 CFENGINE_OPENSSL="$PREFIX/bin/openssl"
 mkdir -p $CFENGINE_MP_DEFAULT_CERT_LOCATION
 mkdir -p $CFENGINE_MP_DEFAULT_KEY_LOCATION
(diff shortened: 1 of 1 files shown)


pasinskim:
trigger build!

pasinskim:
@kacfengine build is green and link is there as expected.

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -233,15 +233,25 @@ done
...
 CFENGINE_MP_KEY=$CFENGINE_MP_DEFAULT_KEY_LOCATION/$CFENGINE_LOCALHOST.key
 if [ ! -f $CFENGINE_MP_CERT ]; then
   $CFENGINE_OPENSSL req -new -newkey rsa:2048 -days 3650 -nodes -x509 -utf8 -sha256 -subj "/CN=$CFENGINE_LOCALHOST" -keyout $CFENGINE_MP_KEY  -out $CFENGINE_MP_CERT -config $PREFIX/ssl/openssl.cnf
+  ln -sf $CFENGINE_MP_CERT $CFENGINE_MP_CERT_LINK
+fi
+
+#
+# If we are upgrading and the link is not there make sure to create it
+#
+if [ ! -f $CFENGINE_MP_CERT_LINK ]; then
```
kacfengine: ``-f`` is a test specifically for *regular* file. You should use ``-h`` instead, which tests for a symbolic link instead.

Most probably this will work regardless, since you will just recreate the link every time, but better to be correct all the way.

kacfengine:
Other than the one comment, it looks good. I don't think you need to run it all the way through the PR analyzer to change that one flag (it's more or less guaranteed to be green), just change it and merge, and make sure you test the package afterwards to be sure.

pasinskim:
@kacfengine thanks for spotting this one. Fixed!

PR #102, opened 2015-11-23
Fix link MP certificate location.

pasinskim:
This is for allowing downloading CSV reports in MP under HTTPS.


pasinskim:
trigger build

pasinskim:
Working as expected. Tested with newest package.

PR #106, opened 2015-11-25
Fix libcurl compile by making sure it honors LDFLAGS.

kacfengine:
Otherwise the Windows 32-bit package does not install. To do this we
must also work around libtool not passing -static-* options to the
compiler like it should by using the CC variable.


kacfengine:
trigger build

kacfengine:
trigger build

PR #107, opened 2015-12-09
Source the proper file to have the needed environment variables for running projects_to_test.

jimis:


jimis:
trigger build

jimis:
trigger build

kacf:
+1

jimis:
Seems to work in debug builds, but must be tested in release builds as well. Deferring that...

PR #113, opened 2016-03-29
Change: Switch SSL Key Size back to 2048

nickanderson:
4096 bit key is overkill. It doesn't offer double the security, but it is
costs computationally more than double. See here:
https://www.gnupg.org/faq/gnupg-faq.html#please_use_ecc


nickanderson:
@jimis & @estenberg

jimis:
+1 from my point of view. @estenberg lets prioritise supporting ECC keys in CFEngine (both MP and agent) instead.

nickanderson:
FYI I also updated the PR to masterfiles to mustachify the httpd config since I added a bundle to generate the cert as well.

nickanderson:
@estenberg what's the verdict here, shall we switch back to 2048 bit ssl keys?

nickanderson:
ping @estenberg

estenberg:
I would just leave it to 4096. I don't think there are any resource issues with this and it takes us above 128 bits of security which makes AES the weakest link not RSA.

Sure ECC would be nice but that is a bigger undertaking entirely and why do we need more bits of security for this?

nickanderson:
OK then closing this PR.

PR #114, opened 2016-03-30
Fix permissions for ppkeys directory in the RHEL7 package.

jimis:


jimis:
Closing because RHEL7 package works after all, as per @GregorioDiStefano 's tests.

PR #117, opened 2016-04-29
CFE-2359: Do not claim to start or stop CFEngine unless it's true.

kacf:
The script figures out whether it needs to start or stop anything, so
before we call it, we don't actually know whether we should output
such a message.

Fix by refactoring cf_console to do general redirection for all
commands, instead of just "echo". This way we can call the script with
messages directed to the console, instead of the log, where all the
rest of the message go.


kacf:
trigger build

kacf:
Tested and fixes the problem!

kacf:

PR #122, opened 2016-05-03
ENT-2724: Fix crash in PHP cleanup function.

kacf:
Both Curl and OpenSSL want to call this function, but it's only
allowed to call it once. This is a bug in PHP itself, so remove for
now. A memory leak in the cleanup is insignificant compared to a
crash.


kacf:
trigger build

pasinskim:
@kacf I don't like the idea of hosting PHP patches, but there is not that much we can do right now. I've opened PHP bug here: https://bugs.php.net/bug.php?id=72140

The change looks safe as this is called in cleanup function, where you are exiting from the thread anyway. Any memory should be freed by OS.

kacf:
Thanks for posting the bug. I agree, it's crappy to maintain our own patches, but it's the best we can do now. I hope they fix it properly upstream.

kacf:

PR #124, opened 2016-05-06
Fix error in package scripts when trying to redirect output on HP-UX.

kacf:
/var/log doesn't always exist on HP-UX.


GregorioDiStefano:
Ha, yes! This explains it :)

PR #127, opened 2016-05-26
ENT-2714 fix libltdl dependency

jimis:
Accidental installation of libtool and related packages on our buildslaves led to our hub package gaining one dependency. This is a series of commits that ensures this won't happen in the future.

### DIFF
--- a/build-scripts/build-environment-check
+++ b/build-scripts/build-environment-check
@@ -1,36 +1,74 @@
-#!/bin/sh -x
+#!/bin/sh
 
 . `dirname "$0"`/functions
 . detect-environment
 . compile-options
 
+
 case "$OS" in
     rhel|centos)
         # Fakeroot is here: http://dl.atrpms.net/el5-$1/atrpms/stable/fakeroot-1.6.4-15.1.el5.$1.rpm
-        DEP_LIST="ntp rsync gcc make bison flex fakeroot gcc-c++ ncurses ncurses-devel pkgconfig rpm-build autoconf pam-devel"
-        CHECK_CMD="rpm -q"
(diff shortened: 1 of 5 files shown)


jimis:
@kacf in the end I implemented removal of unwanted packages, because it's the only way to maintain all these buildslaves from buildscripts. And later I found that `install-dependencies` is actually doing system package removal on some OSes. So maybe I'll consolidate all package removals in one file if I get some time.

kacf:
Remind to never ever run this on my own machine again. :-) But yeah, I guess we need a quick fix for this.

Regarding removing `set -x` and `sh -x`, these are incredibly useful in Jenkins logs, are you sure you want to remove them?

jimis:
On Fri, May 27, 2016 at 8:53 AM, Kristian Amlie <notifications@github.com>
wrote:

> Remind to never ever run this on my own machine again. :-) But yeah, I
> guess we need a quick fix for this.
>

The scripts were already removing packages in other parts; I'll try to
consolidate them into one.


> Regarding removing set -x and sh -x, these are incredibly useful in
> Jenkins logs, are you sure you want to remove them?
>

They are, but they are also extremely confusing, and are difficult to
control when they are hidden all over the source files, inside subshells,
in files we source, etc. The best would only if they were at the top of
scripts, so that they can be turned on and off at will.

That was my line of thought at least, I can bring back the ones I removed
but I think they are minor.

kacf:
> The best would only if they were at the top of scripts, so that they can be turned on and off at will.

Fully agree, but removing it from the top of the script is exactly what you did! :-)

jimis:
Hmmm I guess you are right in the one case. Maybe I did it because I implemented sane logging messages! ;-)

jimis:
Ugh! The script `build-environment-check` is only called by build-remote, and is not used in Jenkins!

PR #128, opened 2016-05-30
ENT-2714 fix libltdl dependency

jimis:


build-scripts/build-environment-check
```diff
@@ -42,16 +42,16 @@ esac
...
 
 RET=0
 for unwanted in $UNWANTED_DEPS
 do
     if query_pkg $unwanted
     then
-        echo "Removing unwanted package:" $unwanted
-        uninstall_pkg $unwanted    ||    { echo FAILURE ; RET=1 ; }
+        echo "ERROR Found unwanted package:" $unwanted
+        RET=1
```
jimis: @kacf do you approve of this commit? ;-)

kacf:
> Portability fix: Exclamation for NOT is not in POSIX.

Where have you read this? It works on every platform we have, including very dumb shells. I know that exclamation point inside a shell test is not POSIX, but outside it should be, IOW `! [ -f /somefile ]` should work, `[ ! -f /somefile ]` will only work in Bash.

jimis:
> Where have you read this? It works on every platform we have, including very dumb shells. I know that exclamation point inside a shell test is not POSIX, but outside it should be, IOW `! [ -f /somefile ]` should work, `[ ! -f /somefile ]` will only work in Bash.

I've experienced exactly the opposite. And googling this indeed shows that exclamation is in POSIX `test`'s [specification](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html). And [shell portable programming manuals](http://www.delorie.com/gnu/docs/autoconf/autoconf_122.html) suggest not to use exclamation outside of test:

>  You may use `!` with test, but not with if: `test ! -r foo || exit 1`.

kacf:
> I've experienced exactly the opposite

Uhm, really...? Because I have experienced the opposite of your opposite... Starting to doubt my own memory now...

Anyway, just try and see what works, but keep an eye out.

PR #132, opened 2016-06-13
Change: Reduce php info leak

nickanderson:
Changelog: Title


nickanderson:
Can we pick this back into 3.9.x and 3.7.x?

nickanderson:
Oh I see its in 3.9.x and 3.7.x already. Thanks!

PR #142, opened 2016-07-29
sanitize spec files

jimis:


jimis:
@nickanderson please review

PR #143, opened 2016-08-10
Change: Use fewer pipes to extract total memory

nickanderson:
awk can find the appropriate line without using cat and grep.


nickanderson:
@maciejmrowiec and or @jimis

kacf:
Both of them are away right now. I assume you have tested the expression (particularly that awk supports `+` and `\s`)?

nickanderson:
Yes, at least on my box I have tested it.


```
> $ cat /proc/meminfo |grep "^MemTotal:.*[0-9]\+ kB"|awk '{print $2}'                                                                  
32824060

> $ awk '/^MemTotal:.*[0-9]+\skB/ {print $2}' /proc/meminfo                                                                            
32824060
```

PR #148, opened 2016-09-02
Switch from OS based to environment based detection of package role.

kacf:
This fits Jenkins better, since we may use the same platform for
multiple roles.

Also, the hub will build on any modern Linux, so just assume that role
on any Linux platform when the Jenkins label is not in the
environment.


kacf:

PR #149, opened 2016-09-05
Adapt to newer Linux distros having moved lock files to /run.

kacf:
We need to include them in the test chroot.


kacf:

PR #150, opened 2016-09-06
Export entire environment to testing chroot, instead of only select o…

kacf:
…nes.


kacf:

PR #151, opened 2016-09-06
Fix CentOS version detection.

kacf:
Specifically for strings like:
'CentOS release 4.5 (Final)'


kacf:

PR #152, opened 2016-09-07
When exporting environment to chroot, treat PATH specially.

kacf:
It needs to add to the existing PATH, not just add it blindly. The
reason is that root has a different path than a user, and we will
need entries from both.


kacf:

PR #156, opened 2016-09-22
Detect cross compilation for Windows using Jenkins labels.

kacf:
If CROSS_TARGET is already set in the environment it will take
precedence though.


kacf:

PR #158, opened 2016-09-23
Work around too long path lengths when packaging Windows MSI.

kacf:
Wix tools have a ridiculously short maximum file length of 128
characters.  This is easily exceeded when Jenkins workspace paths are
involved. Luckily we can shortcut the paths by using Windows drive
letters to point to a directory closer to the source files. Wine will
automatically use the shortest one.


kacf:

PR #160, opened 2016-09-29
ENT-2760: Enterprise: Fix Postgres database migration from 3.8- to 3.9+.

kacf:
The problem was that the initial population of the new database was
run before re-import of the database that was dumped from the old one.
The import script was not prepared for the database to already have
content. However the population scripts are prepared for this, since
this is what would happen if you upgraded between CFEngine versions
that run the same PostgreSQL version.

So the fix is to swap the order, run the import first. However, we
must take care to create the actual databases as well as the database
users first, since these are not covered by the import script.

Fix a couple of other minor things as well.


kacf:
@maciejmrowiec: Particularly the last commit could use a review, to make sure I got the ordering right. I have tested this and works for me; if I create a widget it persists in the upgraded Mission Portal.

kacf:
@nickanderson: Just FYI about the reverted commit, a6abec6.

kacf:

PR #163, opened 2016-10-04
Fetch correct version from 'configure' even for new style AC_INIT.

kacf:
When using AC_INIT to specify the version, the result in 'configure'
is slightly different, now having quotes around it.

### DIFF
--- a/build-scripts/version
+++ b/build-scripts/version
@@ -9,7 +9,11 @@ parse_version_string()
 
   VERSION=$(awk '/^ ?'$VARNAME'=/ {print}' < $BASEDIR/$VERSION_PATH/configure)
 
+  # Remove "VERSION="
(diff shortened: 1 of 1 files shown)


build-scripts/version
```diff
@@ -9,7 +9,11 @@ parse_version_string()
 
   VERSION=$(awk '/^ ?'$VARNAME'=/ {print}' < $BASEDIR/$VERSION_PATH/configure)
 
+  # Remove "VERSION="
   VERSION=${VERSION#*=}
+  # Remove quotes, if any.
+  VERSION=${VERSION%\'}
+  VERSION=${VERSION#\'}
```
jimis: -1 whenever I see this syntax
+1 for the comments
= 0

PR #164, opened 2016-10-04
Get version string properly, even if it's quoted.

jimis:


kacf:
Note that there is a much smaller fix for the quote problem here: https://github.com/cfengine/buildscripts/pull/163

kacf:
The reason it was introduced is because of the change of using `AC_INIT`, but that was already reverted on master.

jimis:
> Note that there is a much smaller fix for the quote problem here: #163

you're using bash magic again, I'll have to say mine is better! ;-)

jimis:
It's [green|https://ci.cfengine.com/job/testing-enterprise-pr/11/] so merging.

jimis:
@kacf I don't specifically enable parallel test harness in this pull request, and I couldn't find any of your code *disabling* parallel test harness. Where did you do that?

kacf:
Parallel test harness is the default in newer autoconf. Search for `serial-tests` and `SERIAL_TESTS` in `configure.ac`.

PR #170, opened 2016-10-14
INF-370: Fix Community jobs building Enterprise packages

jimis:
Because all packages-* jobs come from the same upsteam job, they all have the "nova" directory. Thus auto-detection was building Enterprise packages everywhere.

Fixed by looking for a "JOB_NAME" environment variable that is usually
set by Jenkins.


jimis:
Thanks for the review @nickanderson !

PR #173, opened 2016-10-28
ENT-2846 Enable cf-serverd unit by default

nickanderson:


kacf:
Cherry-picked to 3.10.x as well.

PR #175, opened 2016-11-02
Solaris: Work around faulty detection of 32-bit userland on 64-bit ke…

kacf:
…rnel.


kacf:
Cherry-picked to 3.7.x and 3.10.x as well.

PR #176, opened 2016-11-03
ENT-2846 Fix permission errors importing SQL

nickanderson:


nickanderson:
ping @jimis and or @kacf any issues here? should we be setting the perms back after the data import ?

kacf:
* Ticket reference?
* Why those scripts in particular? There are other scripts that don't get this treatment, what about them?
* I'm thinking it's better to enable read access to everyone, but there is no ticket reference here, so I don't know the exact issue.

nickanderson:
Sorry, it was only in the branch name. it's ENT-2846

kacf:
Ok. I still think this isn't the right fix. It only covers some files, not all of them, for no reason that I can see, and enabling `chmod go+r` seems like a less intrusive fix to me.

kacf:
After discussion in the ticket, I think we can accept this. It's not an ideal solution, but fixing it properly would be a much larger and riskier fix.

Have you tested the resulting package?

nickanderson:
@kacf I just kicked off pr build 78 with a few minor changes, i will test that.

packaging/cfengine-nova-hub/debian/rules
```diff
@@ -83,6 +83,13 @@ install: build
 	cp $(BASEDIR)/nova/db/schema_settings.sql $(CURDIR)/debian/tmp$(PREFIX)/share/db/
 	cp $(BASEDIR)/nova/db/ootb_settings.sql $(CURDIR)/debian/tmp$(PREFIX)/share/db/
 
+# ENT-2708, ENT-2846
+	chmod 600 $(PREFIX)/share/GUI/application/config/*.php
+	chmod 600 $(PREFIX)/share/GUI/phpcfenginenova/*.sql
+	chmod 700 $(PREFIX)/share/GUI/phpcfenginenova/migrations
+	chmod 600 $(PREFIX)/share/GUI/phpcfenginenova/migrations/*.sql
+	chmod 600 $(PREFIX)/share/db/*.sql
```
nickanderson: I guess this should have happened for $(CURDIR)/debian/tmp$(PREFIX) instead @jimis?

jimis: I /guess/ by reading the context, but better ask @kacf who works more Debian-based platforms.

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -384,10 +389,19 @@ else
...
 
+
+  # Return restrictive permissions to sql files ENT-2684
+  (chown root:root "$PREFIX/share/GUI/phpcfenginenova/*.sql")
+
   #create database for hub internal data
   (cd /tmp && su cfpostgres -c "$PREFIX/bin/psql cfsettings -f $PREFIX/share/db/schema_settings.sql")
   (cd /tmp && su cfpostgres -c "$PREFIX/bin/psql cfsettings -f $PREFIX/share/db/ootb_settings.sql")
 
+  (cd /tmp && su cfpostgres -c "$PREFIX/bin/psql cfdb -f $PREFIX/share/db/schema.sql")
```
kacf: This is a brand new SQL import line. Did you mean to include this? If so, I'd like a comment about this in the commit message.

kacf: > Ref: https://github.com/cfengine/buildscripts/commit/dccaeb4d282026e61bb745455367ccf652f14f4b#diff-6570e7c66339d9b71de2575e972f19eaR373

Yes, but the line already exists further up. You've added it a second time.

nickanderson: Not sure how it got in there, will redact.

PR #190, opened 2016-12-19
Do not write pkg-store destination file if already existing

jimis:


jimis:
Green in Jenkins, merging.

PR #191, opened 2016-12-20
Enable parallel plans in PG 9.6+

maciejmrowiec:
Signed-off-by: Maciej Mrowiec <mrowiec.maciej@gmail.com>

safe for 3.10.x+


kacf:
Should be merged post 3.10.0 IMHO. We've already spent a lot of time on testing 3.10.0, so should be blockers only.

kacf:
Cherry-picked to 3.10.x as well.

PR #193, opened 2016-12-21
pkg-cache list: do not print literal '*' if the directory is empty.

jimis:
Also improved some logging messages.


jimis:
This should fix another spurious error that I noticed, when the cache directory was empty.

PR #206, opened 2017-02-21
INF-411: 3.7.x cherry-picks

jimis:


jimis:
Jenkins is green at last, merging!

PR #212, opened 2017-03-17
ENT-2758: Reloacte Mission Portal application logs

nickanderson:
Mission Portals application logs have moved from
/var/cfengine/httpd/htdocs/application/logs to
/var/cfengine/httpd/logs/application

Changelog: Commit


jimis:
looks good to me

PR #216, opened 2017-04-21
ENT-3129: Run database migrations during post-install

nickanderson:
This runs database migrations during post-install. Note, it always tries
to move forward (even though migrations support going backwards with
the "down" option). If a package is either upgraded or downgraded then
the migration runs. We don't explicitly support downgrading packages
currently so I believe this is OK even if not ideal.

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -441,6 +441,9 @@ EOF
 
 fi
 
+# Proper Database Migrations
+$PREFIX/httpd/php/bin/php $PREFIX/httpd/htdocs/index.php migration/tasks/migrate
+
 #
 # Apache related
 #


nickanderson:
Needs to be merged with https://github.com/cfengine/mission-portal/pull/320

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -441,6 +441,9 @@ EOF
 
 fi
 
+# Proper Database Migrations
+$PREFIX/httpd/php/bin/php $PREFIX/httpd/htdocs/index.php migration/tasks/migrate
```
kacf: Should this be surrounded by `if is_upgrade; then`?

nickanderson: Well, on a fresh install of the package I had to run the migration for it to work, so if this is guarded to only when upgrade, a new OOTB install won't work.

kacf: Alright, disregard my comment then.

PR #218, opened 2017-05-05
ENT-3164: Update dependencies for 3.11 Beta

Lex-2008:


jimis:
Please add a "changelog" tag to the removed deps commit message.

deps-packaging/lcov/cfbuild-lcov.spec
```diff
@@ -24,7 +24,7 @@ exit 0
 
 %install
 rm -rf $RPM_BUILD_ROOT
-make install PREFIX=$RPM_BUILD_ROOT
+make install DESTDIR=$RPM_BUILD_ROOT PREFIX=/usr CFG_DIR=/etc
```
jimis: Why was this change needed? Add the reason to the commit message.

Lex-2008: "Copy `make install` command from build script provided in the package" sounds good?

Also, I squashed together commits for changing version and making necessary changes in our buildscripts. Do you want me to unsquash them back?

build-scripts/compile-options
```diff
@@ -3,7 +3,6 @@
 # Calculate dependencies and build parameters according to the environment.
 # SOURCE this file, don't execute.
 
-
```
jimis: Can you add an explanation of why we removed the dependency in the commit message? (It doesn't have to be in the changelog.) For reference in case somebody tries to bring back the dependency in the future. Also mention the ticket number.

build-scripts/compile-options
```diff
@@ -171,7 +171,7 @@ case "$ROLE" in
         ;;
     # AGENT-ONLY dependencies
     agent)
-        var_append DEPS "libcurl postgresql"
```
jimis: Reason for removal in the commit message, and mention the ticket number.

deps-packaging/lcov/cfbuild-lcov.spec
jimis: Explanation needs to show the root cause, like "upstream changed build instructions". No need to unsquash.

jimis:
Nice, thank you @Lex-2008 !

What do you think is best? Cherry-picking the relevant commits to 3.10.x/3.7.x now, or waiting for the release of these branches and updating dependencies then?

Lex-2008:
I think we do only minor version updates on that branches, so, for example, 3.10 will always be on git 2.10.x branch, while 3.11 is on git 2.13.0 now, after this commit. Hence, these commits are not relevant to those branches.

jimis:
Almost. But while we try to keep breakage to a minimum in LTS branches, we would still upgrade the major version of a dependency if deemed important, if for example the old version of the dependency imposed a security risk.

Lex-2008:
Hm, then we need to investigate it on a case-by-case basis. Let's have a look at it when we come to upgrading relevant 3.7/3.11 branches

jimis:
On Tue, May 16, 2017 at 11:26 AM, Aleksei Shpakovskii <
notifications@github.com> wrote:

> Hm, then we need to investigate it on a case-by-case basis. Let's have a
> look at it when we come to upgrading relevant 3.7/3.11 branches


OK!

PR #219, opened 2017-05-12
INF-416: Rewrite pkg-cache to use SFTP instead of nfs mounted local directories

jimis:
Care has been taken to atomically move directories around in order to
avoid race conditions when 2 buildslaves are uploading the same files,
or when one is uploading and one is downloading.


jimis:
@Lex-2008 @nickanderson I'm ready to merge, /maybe/ you want to have a look.

PR #225, opened 2017-06-09
ENT-3205, ENT-3218: Warn that RPM hub package upgrade is unsupported

jimis:
After package upgrade a bunch of old and new files are mixed up in
htdocs directory, for example old and new Code Igniter versions coexist.
Unfortunately it's not trivial to just delete the old files, because
they are manually copied after package installation, and workarounds are
messy because of the mess the hub package currently is (see ENT-3218).


jimis:
Mostly green, merging.

PR #229, opened 2017-07-12
unshallow repos

Lex-2008:
Travis checks out with `--depth` parameter[1], which implies
`--single-branch`[2]. This confuses determine-version.py script, so we
need to convert the repo to normal one.

[1]: https://docs.travis-ci.com/user/customizing-the-build#Git-Clone-Depth

[2]: https://git-scm.com/docs/git-clone#git-clone---depthltdepthgt


Lex-2008:
Right, it's not that easy...

PR #230, opened 2017-07-13
checkout branches instead of "unshallow" operation

Lex-2008:
since new determine-version.py requires them

Maybe we should edit relevant script in core, too

### DIFF
--- a/build-scripts/travis
+++ b/build-scripts/travis
@@ -15,7 +15,14 @@ do
         then
             (
             cd $i
-            git fetch origin refs/heads/*:refs/remotes/origin/*
+            # Rename the branch we're on, so that it's not in the way for the
+            # subsequent fetch. It's ok if this fails, it just means we're not on any
+            # branch.
+            git branch -m temp-branch || true
+            # Fetch rest of branches
+            git fetch origin 'refs/heads/*:refs/remotes/origin/*'
+            # Get last remaining tags, if any.
+            git fetch --tags origin
             )
         fi
     else
--- a/build-scripts/travis
+++ b/build-scripts/travis
@@ -23,6 +23,8 @@ do
             git fetch origin 'refs/heads/*:refs/remotes/origin/*'
             # Get last remaining tags, if any.
             git fetch --tags origin
+	    # This was partially taken from Mender repo:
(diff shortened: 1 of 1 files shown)


build-scripts/travis
```diff
@@ -1,18 +1,26 @@
...
+        # ./3rdparty/core/determine-version.py file exists.
+        # (Other repos are not checked for versions, we assume).
+        if [ "$i" = "core" -o -f $i/3rdparty/core/determine-version.py ]
+        then
+            (
+            cd $i
+            git fetch origin refs/heads/*:refs/remotes/origin/*
+            )
+        fi
+    else
```
kacf: I recommend [this one](https://github.com/mendersoftware/mender/blob/master/.travis.yml#L61-L69) that we use in Mender. It has been proved and tested for a long time already. The reason for the branch trickery is that branch builds, tag builds and pull request builds all have subtle differences.

Lex-2008: Branch renaming looks interesting, will add!

"Fetching directly into local branches" is not really needed to fix anything, right? The thing is that currently `determine-version.py` looks only at remote branches starting with "origin" or "upstream", and probably we'd like to keep it like this.

Lex-2008: Do you know if I should add a comment pointing to Mender repo where this was taken from? :-)

kacf: > Do you know if I should add a comment pointing to Mender repo where this was taken from? :-)

A comment is always helpful! :-)

> "Fetching directly into local branches" is not really needed to fix anything, right? The thing is that currently determine-version.py looks only at remote branches starting with "origin" or "upstream", and probably we'd like to keep it like this.

Right, I think this was done in order to guarantee that no branch name was already taken. Not sure how important it is. I'm sure fetching to remote branches is fine too, if you need that.

Lex-2008:
@jimis:
Checking out branches - yes, needed - branches are used by `determine-version.py` now - enterprise repo currently fails building in Travis without it (looks like it doesn't see all tags).

Renaming current branch - probably it might help in some cases. Do you want to disable it until it fails? :-)

Fetching tags - might be also needed, since they're also used in `determine-version.py`.

Note that Travis checks out only last 50 commits of current branch. So for most git operations we need the rest of git history.

PR #234, opened 2017-07-18
Switch to http mirror for libcurl

Lex-2008:
Signed-off-by: Aleksei Shpakovskii <aleksei.shpakovskii@cfengine.com>


jimis:
Is this a cherry-pick? If so, always use `git cherry-pick -x` so that the source commit is shown. You can also push cherry-picks directly (without pull requests) if you are confident things will not break.

Lex-2008:
added comment to the commit message, merging

PR #254, opened 2017-09-13
ENT-3284: Run ldap migration script in postinstall.sh

Lex-2008:
Signed-off-by: Aleksei Shpakovskii <aleksei.shpakovskii@cfengine.com>


Lex-2008:
will be done in https://github.com/cfengine/buildscripts/pull/259

PR #259, opened 2017-09-26
All LDAP things combined

Lex-2008:


Lex-2008:
Note: "ENT-3284: Also run postinstall script" should be renamed to "ENT-3382: Run ldap post-install.sh after installation"

PR #263, opened 2017-10-03
CFE-2029: Remove package installation of cf-twin

nickanderson:
Changelog: Title

@jimis


jimis:
Did you grep everywhere for more `cf[-_]twin` occurences?

nickanderson:
No I didn't grep everywhere. I don't think we knowingly use cf-twin any longer. We gotta rip out the remanents eventaully.

PR #264, opened 2017-10-04
ENT-3277: Update travis to use stages.

Lex-2008:
Also provide a simple sample .travis.yml file

Note that this PR will break Travis building in other repos, so it should be merged together with relevant PRs to other repos, and after that all PRs in other repos should be rebased.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+# This is a sample file which should be put into other repos.
+# Idea is that other repos (nova, enterprise, mission-portal) should check out
+# this repo (buildscripts) and run common travis script from this repo
+
+sudo: required
+language: generic
+
+install: cd .. && git clone git@github.com:cfengine/buildscripts.git
+script: buildscripts/build-scripts/travis
+
+jobs:
+  include:
+
+    - stage: build
+      env: JOB=build
+      language: c
+      cache:
+        directories:
+          - $HOME/.cache/
+
+
+    - stage: test
+      env: JOB=acceptance-test
+      if: not type = cron
+
+    - stage: test
+      env: JOB=deployment-test
+      if: not type = cron
--- a/build-scripts/travis
+++ b/build-scripts/travis
@@ -1,85 +1,173 @@
 #!/bin/bash
 
 set -ex
-for i in core nova enterprise mission-portal masterfiles design-center
-do
-    if [ -d $i ]
-    then
-        # Repo already checked out. Probably by Travis.
-        # Travis checks out only one branch, what confuses `determine-version.py`.
-        # To fix it, we need to ask git to checkout other branches, too.
-        # But we do it only for "core" and those repos where
-        # ./3rdparty/core/determine-version.py file exists.
-        # (Other repos are not checked for versions, we assume).
-        if [ "$i" = "core" -o -f $i/3rdparty/core/determine-version.py ]
-        then
-            (
-            cd $i
-            # Rename the branch we're on, so that it's not in the way for the
-            # subsequent fetch. It's ok if this fails, it just means we're not on any
-            # branch.
-            git branch -m temp-branch || true
-            # Fetch rest of branches
-            git fetch origin 'refs/heads/*:refs/remotes/origin/*'
-            # Get last remaining tags, if any.
-            git fetch --tags origin
-	    # This was partially taken from Mender repo:
-	    # https://github.com/mendersoftware/mender/blob/fc27422420bd3859d917579484856debdfa50e5f/.travis.yml#L61-L69
-            )
-        fi
-    else
-        git clone git@github.com:cfengine/$i.git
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,24 @@
+sudo: required
+language: generic
+
+install: cd .. && git clone git@github.com:cfengine/buildscripts.git
```
jimis: why is this needed?

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+      env: JOB=build
+      language: c
+      cache:
+        directories:
+          - $HOME/.cache/
+
+
+    - stage: test
+      env: JOB=acceptance-test
+      if: not type = cron
```
jimis: why not run cron jobs?

build-scripts/travis
```diff
@@ -1,85 +1,139 @@
...
+  echo $ARTIFACTS_KEY
+  echo $ARTIFACTS_SECRET
+  echo 
+  echo 
+  echo 
+  echo 
+  echo n
+  echo y
+) | s3cmd --configure >/dev/null
+
```
jimis: lol

.travis.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    - stage: test
+      env: JOB=deployment-test
```
jimis: +1 for parallelizing the different kinds of tests

build-scripts/travis
```diff
@@ -1,85 +1,139 @@
...
+        s3cmd get $DEB_PACKAGE cfe.deb
+        sudo dpkg -i cfe.deb
+        touch /home/travis/rpmvercmp
+        chmod a+x ~/rpmvercmp
+        cd ..
+        test -d core || git clone git@github.com:cfengine/core.git
+        test -d enterprise || git clone git@github.com:cfengine/enterprise.git
+        chmod -R g-w */tests/acceptance
+        cd core/tests/acceptance
+        rm -rf ./05_processes/01_matching/timed/stime.cf ./05_processes/01_matching/promiser_match_is_correct.cf ./07_packages/00*.cf ./10_files/11_xml_edits ./25_cf-execd
```
jimis: OK for now but create the relevant task to clean-up this.

build-scripts/travis
```diff
@@ -1,85 +1,139 @@
...
+        AGENT=/var/cfengine/bin/cf-agent CF_PROMISES=/var/cfengine/bin/cf-promises CF_SERVERD=/var/cfengine/bin/cf-serverd CF_KEY=/var/cfengine/bin/cf-key CF_EXECD=/var/cfengine/bin/cf-execd CF_NET=/var/cfengine/bin/cf-net CF_RUNAGENT=/var/cfengine/bin/cf-runagent RPMVERCMP=~/rpmvercmp ./testall --tests=common,timed,slow,errorexit,serial,network,libxml2,libcurl,unsafe --gainroot=sudo
```
jimis: Please break the long lines, it's very hard to read otherwise. 78 is the golden number, but feel free to deviate if needed. :-)

.travis.yml
Lex-2008: Probably this should be added as a comment on top:
```
# This is a sample file which should be put into other repos.
# Idea is that other repos (nova, enterprise, mission-portal) should check out
# this repo (buildscripts) and run common travis script from this repo
```
Indeed, this script is not expected to work with this repo.
Primarily, because it will fail on `git clone` step (`pwd` at the beginning of Travis run is something like /.../cfengine/buildscripts/, so after `cd ..` you're in a dir which already has a 'buildscripts' subdir)
Secondary, and that's more important, because buildscripts is a public repo, which doesn't have access to our private repos. This can be fixed, but not right now.

.travis.yml
Lex-2008: I'm thinking about using cron jobs only to refresh cache with dependency packages. Do you think we should run tests daily, too?

build-scripts/travis
Lex-2008: OK, https://tracker.mender.io/browse/CFE-2698 created!

.travis.yml
Lex-2008: Note, however, that CFEngine company is on the cheapest plan currently, so we don't really parallelize anything. Moreover, running two jobs sequentially in two different stages causes old VM to be destroyed and new one to be booted, thus taking time (about few minutes).

build-scripts/travis
Lex-2008: ok I've wrapped some lines and amended last commit

.travis.yml
jimis: Yes. Things some times cause breakage, which is detected after they are already in.

jimis:
Thanks @Lex-2008, oneliners are now readable, feel like art.

PR #275, opened 2017-10-17
ENT-3336: Ensure Enterprise Hub package upgrade aborts when unable to…

Lex-2008:
… backup database

Changelog: Title
(cherry picked from commit 8053ed91989e91c5dc73068fc4fb5a400c9e1dd3)

### DIFF
--- a/packaging/common/cfengine-hub/preinstall.sh
+++ b/packaging/common/cfengine-hub/preinstall.sh
@@ -32,12 +32,15 @@ if is_upgrade && egrep '^3\.[6-9]\.' "$PREFIX/UPGRADED_FROM.txt" >/dev/null && [
   FAILED=0
   for db in $CF_DBS; do
     cf_console echo "Backing up database $db..."
-    (cd /tmp && su cfpostgres -c "$PREFIX/bin/pg_dump $db | gzip -c > $PREFIX/state/pg/db_dump-$db.sql.gz")
+    su cfpostgres -c "$PREFIX/bin/pg_dump $db" > $PREFIX/state/pg/db_dump-$db.sql
+
     if [ $? != 0 ]; then
       FAILED=1
       cf_console echo "Not able to migrate database. Aborting."
       break
     fi
+
+    gzip $PREFIX/state/pg/db_dump-$db.sql
   done
 
   if [ "$FAILED" != 0 ]; then


Lex-2008:
Separated `pg_dump` and `gzip` commands into individual steps.
Checking result of only one of them.
Also not sure why running `pg_dump` as root was considered a bad idea.
Quickly tested - works as expected, both when it can and when it can't backup.

Any objections? @jimis , @nickanderson , @kacf ?

packaging/common/cfengine-hub/preinstall.sh
```diff
@@ -32,12 +32,15 @@ if is_upgrade && egrep '^3\.[6-9]\.' "$PREFIX/UPGRADED_FROM.txt" >/dev/null && [
   FAILED=0
   for db in $CF_DBS; do
     cf_console echo "Backing up database $db..."
-    (cd /tmp && su cfpostgres -c "$PREFIX/bin/pg_dump $db | gzip -c > $PREFIX/state/pg/db_dump-$db.sql.gz")
+    $PREFIX/bin/pg_dump $db  > $PREFIX/state/pg/db_dump-$db.sql
```
kacf: > Also not sure why running pg_dump as root was considered a bad idea.

In this particular case I suspect that root may actually have less privileges than cfpostgres, the reason being that the permissions are enforced by the database, not by the kernel. But it's only a suspicion, maybe check with @maciejmrowiec.

maciejmrowiec: yeh, cfpostgres is the main admin user

packaging/common/cfengine-hub/preinstall.sh
```diff
@@ -32,12 +32,15 @@ if is_upgrade && egrep '^3\.[6-9]\.' "$PREFIX/UPGRADED_FROM.txt" >/dev/null && [
...
+
     if [ $? != 0 ]; then
       FAILED=1
       cf_console echo "Not able to migrate database. Aborting."
       break
     fi
+
+    gzip $PREFIX/state/pg/db_dump-$db.sql
```
maciejmrowiec: pg_dump can compress on the fly

maciejmrowiec: `pg_dump -Fc`

jimis: Unfortunately I doubt this will work since we build PostgreSQL with `--without-zlib`. Don't ask me why, I have no clue. Since we build zlib separately for CFEngine, we should do `--with-zlib=/var/cfengine` instead.

Lex-2008: Right, it's just not `-Fc`, but `-Z9`, and indeed
```
pg_dump: WARNING: requested compression not available in this installation -- archive will be uncompressed
```
I've created a separate issue https://tracker.mender.io/browse/ENT-3441 for that

jimis: Great! But I think `-Fc` is also compressed, and a better format suitable for `pg_restore`:

```
c
custom

    Output a custom-format archive suitable for input into pg_restore. Together with the directory output format, this is the most flexible output format in that it allows manual selection and reordering of archived items during restore. This format is also compressed by default.
```

Not sure if it's transparent to change format for our scripts. Do we ever use this archive that we are storing?

Lex-2008: hm,
> This format is also compressed by default.

Ok, probably, when it's supported. Because currently, `-Fc` takes more space than `-Fp` (default one):
```
~# /var/cfengine/bin/pg_dump -Fc cfdb | wc -c
2067518
~# /var/cfengine/bin/pg_dump -Fp cfdb | wc -c
1742252
```

Regarding restoring - indeed, `-Fc` is a binary format and is suitable for `pg_restore`, while `-Fp` looks like plaintext SQL and is suitable for feeding into `psql`, according to https://www.postgresql.org/docs/9.1/static/app-pgdump.html

As a Linux neophyte, I'm voting for plaintext format.

nickanderson:
+1 for plain text

PR #298, opened 2017-12-21
pack ssl/openssl.cnf, delete some other for openssl

Lex-2008:
ssl/openssl.cnf file is needed for CFEngine and should be also included in openssl packaging.

Other files (that are not needed) should be deleted from there.


Lex-2008:
testbuild: https://ci.cfengine.com/view/Enterprise/job/testing-enterprise-pr/512/

PR #288, opened 2017-11-30
Update dependencies for master

Lex-2008:
Bootstrap job: [Build Status](https://ci.cfengine.com/job/bootstrap-enterprise-pr/464/)
Testing job: [Build Status](https://ci.cfengine.com/job/testing-enterprise-pr/464/)

PR #300, opened 2017-12-21
Fix errors about undeclared curl_off_t on AIX and HP-UX.

jimis:
This patch has been submitted upstream:
https://github.com/curl/curl/pull/2186

The format of the patch file is very basic, because AIX patch(1) does not work
with neither git diffs, context diffs, nor unified diffs.


jimis:
@Lex-2008 this should fix the failing platforms. I brought the buildslaves back online, you can trigger a build with this and merge it if it passes.

Lex-2008:
Thanks!

Lex-2008:
I've incorporated your commit into my PR: https://github.com/cfengine/buildscripts/pull/288 so this might be closed, probably.

PR #303, opened 2018-01-02
312 deps

jimis:


jimis:
Supersedes #288

jimis:
The build is green now.

jimis:
For AIX and HP-UX I reverted back to the fix from @Lex-2008, as upstream might take longer to decide what to merge (see curl/curl#2214 ).

PR #304, opened 2018-01-03
ENT-2746: Tighten postgres socket permissions

nickanderson:
Now the postgres socket permissions are only accessible by cfapache, cfpostgres,
and root.

Changelog: Title

### DIFF
--- a/packaging/common/cfengine-hub/preinstall.sh
+++ b/packaging/common/cfengine-hub/preinstall.sh
@@ -194,15 +194,22 @@ then
   fi
 fi
 #
-# We need a cfapache user for our web server
+# We need a cfapache user and group for our web server
 #
 /usr/bin/getent passwd cfapache >/dev/null || /usr/sbin/useradd -M -r cfapache
 /usr/bin/getent group cfapache >/dev/null || /usr/sbin/groupadd -r cfapache
 
 #
-# We check if there is a postgres user already, otherwise we create one
+# We make sure there is a cfpostgres user and group
 #
 /usr/bin/getent passwd cfpostgres >/dev/null || /usr/sbin/useradd -M -r cfpostgres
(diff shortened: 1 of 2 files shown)


packaging/common/cfengine-hub/preinstall.sh
```diff
@@ -203,6 +203,13 @@ fi
 # We check if there is a postgres user already, otherwise we create one
 #
 /usr/bin/getent passwd cfpostgres >/dev/null || /usr/sbin/useradd -M -r cfpostgres
+/usr/bin/getent group cfpostgres >/dev/null || /usr/sbin/groupadd -r cfpostgres
+
+#
+# We make sure that the cfapache user is part of the cfpostgres group so that
+# the webserver can read from the socket ENT-2746
+#
+getent group cfpostgres | grep &>/dev/null "\b${cfapache}\b" || /bin/gpasswd --add cfapache cfpostgres
```
jimis: I can't read this line. What is the grep line doing? don't use extended regexes for simple stuff. don't getent again, you've done it before, group should be there. Redirection should be after arguments, for readility sake.

Otherwise this pull request looks good.

nickanderson: The first `getent` is to see if the group exists, if it does not then add the group. The second `getent` and grep are to see if the `cfapache` user is in the `cfpostgres` group. If it is not,  then add the `cfapache` user to the `cfpostgres` group.

I changed the grep to be less precise.
I moved the redirection after the search string.

jimis:
Alright thanks for explaining. BTW I did not ask you to change the regex to an imprecise one, but to a standard one. What is that backslash-b you were doing? `man 7 regex` does not mention it. Personally I wouldn't even grep for it, just check the return value of `getent`.

nickanderson:
`\b` is word boundary in PCRE.
I can change it to simpley getentgroup && gpasswd --add

nickanderson:
Rebased and squashed.

PR #323, opened 2018-02-08
ENT-3217: nicer httpd/htdocs clean up

Lex-2008:
share/GUI dir is managed by package manager, while httpd/htdocs is
populated by MPF with some files later added by PHP scripts.

Before upgrading, we remove all files from httpd/htdocs dir, which have
corresponding files in share/GUI.

If share/GUI dir is not found, but httpd/htdocs exists (it shouldn't
normally happen), we fallback to old method of purging all files except
known exceptions.

### DIFF
--- a/packaging/common/cfengine-hub/preinstall.sh
+++ b/packaging/common/cfengine-hub/preinstall.sh
@@ -1,14 +1,4 @@
 
-if is_upgrade
-then
-    cf_console echo  \
-        'WARNING: attempted hub upgrade detected;'  \
-        'this is UNSUPPORTED, aborting!'            \
-        '(relevant tickets: ENT-3205, ENT-3218)'
-    # exit 1
-fi
-
-
 if is_upgrade; then
   # This is nice to know to provide fixes for bugs in already released
   # package scripts.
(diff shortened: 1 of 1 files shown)


nickanderson:
Mmmmmkay the code ignitor upgrade renamed many files. Those left behind files were problematic. Purging only files that exist in share GUI wouldn't cover this case would it?

Lex-2008:
I assume that CI upgrade renamed files in `share/GUI` first, right?
`preinstall.sh` script runs when `share/GUI` still has old CI files, so it should work like this:

* first we have old CI both in `share/GUI` and `httpd/htdocs`
* then `preinstall.sh` wipes old CI from `httpd/htdocs`, leaving there only `cf_robot.php`
* then package manager replaces old CI with new one in `share/GUI`
* then MPF copies new CI from `share/GUI` to `httpd/htdocs`

Am I wrong somewhere?

nickanderson:
I don't have the full understanding of the order of the packaging script activations across different systems and how some of our scripts are called more than once. I know there is a matrix to reference. It sounds like you have it all in your head so I will defer to your judgment here. All I care is that it works in the end.

Lex-2008:
There is a good overview in our repo: https://github.com/cfengine/buildscripts/blob/master/packaging/common/README-install-scripts.txt#L39
TL;DR: "preinstall" script runs before upgrade and installation and hence is a proper place to clean up before installation.
Yes, it works, as far as deb/rpm systems are concerned, and we don't make hubs for other systems :-)

nickanderson:
OK, merge when your ready.

PR #331, opened 2018-02-20
Increase the level of parallelism to 2 * NUM_CPUS

vpodzime:
All our operations are IO-bound so having twice the number of
jobs as CPUs is not a problem. The rare cases (if there are even
any) where this makes things slightly slower due to
process/context switching are well worth the extra speed.

Also make sure a temporary value for the MAKEFLAGS variable is
really temporary.

### DIFF
--- a/build-scripts/detect-environment
+++ b/build-scripts/detect-environment
@@ -328,7 +328,7 @@ detect_cores()
 
   # Make number of jobs one higher than core count, to account for I/O, network, etc.
   echo "Detected amount of CPU cores is $NUM_CORES"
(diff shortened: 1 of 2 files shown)


vpodzime:
Unfortunately, this has a negligible impact on the time the tests run for because they already run in parallel batches using different mechanisms than just the ``MAKEFLAGS``.

olehermanse:
@vpodzime should this be closed?

vpodzime:
> @vpodzime should this be closed?

Yup.

PR #332, opened 2018-02-21
ENT-3247: Remove old LDAP. Add packaging rules for including ldap-api…

craigcomstock:
… module as htdocs/ldap in nova-master and nova-cp builds.


craigcomstock:
And now this is really just a change for local builds I guess since jenkins builds don't use build-remote and jenkins config already checks out ldap-api to mission-portal/ldap. Thanks @Lex-2008 for the help.

PR #341, opened 2018-03-22
Attempting to fix pgtap

olehermanse:


Lex-2008:
thanks, but some more work was needed. I resolved the issue in some other PRs.

PR #346, opened 2018-03-28
Attempt to fix the windows packages

vpodzime:
We will see how far and deep this goes eventually.


craigcomstock:
already merged a different fix at https://github.com/cfengine/buildscripts/pull/349

PR #350, opened 2018-04-02
try to get travis running "sudo make check" for nova reporting t…

craigcomstock:
…ests


craigcomstock:
@olehermanse @Lex-2008 take a look at this one and https://github.com/cfengine/nova/pull/1126 let's merge them both and see if travis can succeed.

PR #351, opened 2018-04-03
ENT-3691 fixup TEST_OPTIONS export line so that it works properly.

craigcomstock:
@Lex-2008 please review and merge ;)


craigcomstock:
nevermind @Lex-2008, @nickanderson gave me right writes.

PR #353, opened 2018-04-03
ENT-3705: need to run autogen before doing "sudo make check" for nova…

craigcomstock:
… reporting tests


craigcomstock:
@Lex-2008 is it just me or does it seem silly to have all the logic that runs reporting tests in buildscripts when it only runs for nova... so why not have it in .travis.yml? Just a thought. ;) Would make testing a PR easier since no dep on buildscripts changes.

PR #354, opened 2018-04-04
ENT-3525 fix reporting tests execution for travis

craigcomstock:
@Lex-2008 turns out there is a perfectly suitable and easily available method for running the .test files: /usr/bin/prove! (provided by Perl where TAP format tests came from in the first place).

I ran through a debug session on travis and all seemed fine with this change.


Lex-2008:
These little gems of UNIX... When one day you learn "oh, nice program! Wow, my machine already has it installed!"

PR #361, opened 2018-04-17
3.10.x deps update including libxml2 huge_val patch

craigcomstock:

PR #362, opened 2018-04-19
Fix a few sources URLs on 3.10.x

vpodzime:
So that the download URL works again.


vpodzime:
This should never get in. #361 is the way to go.

vpodzime:
This is just a temporary change to make builds of https://github.com/cfengine/core/pull/3089 possible.

PR #365, opened 2018-04-23
Install python-psycopg2 for reporting tests

vpodzime:
The new tests written in python need this package to do SQL
queries on PostgreSQL DBs.


Lex-2008:
I am 21 minutes late... Yes, it's good!

PR #366, opened 2018-04-25
libxml

Lex-2008:


vpodzime:
The failure with the *.patch* file owner/group can be fixed with this:
```
diff --git a/deps-packaging/pkg-build-rpm b/deps-packaging/pkg-build-rpm
index bc92b9c..2a666b5 100755
--- a/deps-packaging/pkg-build-rpm
+++ b/deps-packaging/pkg-build-rpm
@@ -69,8 +69,9 @@ then
         ln -sf $srcfile $BASEDIR/$PKGNAME/SOURCES
     done
 
-    # Almost, but not completely, equivalent to -a, compatible with AIX.
     cp -pPR $P/* $BASEDIR/$PKGNAME/SOURCES
+    # make sure the SOURCES are owned by a known user ('root' is always known)
+    sudo chown -R root:root $BASEDIR/$PKGNAME/SOURCES
 fi
```
But then it hits some other error, this time with an invalid version string -- see https://ci.cfengine.com/job/testing-pr/421/label=PACKAGES_HUB_x86_64_linux_redhat_5/console

PR #369, opened 2018-04-26
ENT-3761: Fix php build for rhel5 due to isfinite reference

craigcomstock:
option #1, don't upgrade php, I don't recommend it but in case nova/mission portal complain.


craigcomstock:
deprecated

PR #370, opened 2018-04-26
ENT-3761: Upgrade php 7.2.5 and fix rhel5 php problems

craigcomstock:
option #2, upgrade php and fix rhel5 build problem for php


craigcomstock:
Odd. I thought I had put it there. Will fix.

craigcomstock:
Ok. Squashed and will test again before merging.

craigcomstock:
just to be safe, I'm going to only fix the issue and not upgrade php.

PR #373, opened 2018-05-08
Travis: Added a deployment test with SQL asserts

olehermanse:


olehermanse:
https://github.com/cfengine/nova/pull/1178

PR #375, opened 2018-05-09
Add files specifying which labels to run jenkins jobs on

vpodzime:
These branch-specific files define where to run the jenkins jobs
on by default and when running on exotic platforms too.


vpodzime:
Example runs:

vpodzime:
I cannot really integrate this with jenkins before it's merged because that would break the jenkins jobs for other PRs and branches that don't have these files. Also, this needs to be backported to 3.7.x and 3.10.x.

build-scripts/get_labels_expr.py
```diff
@@ -0,0 +1,54 @@
...
+from __future__ import print_function
+
+import argparse
+import sys
+import re
+
+COMMENT_OUT = re.compile(r"([^#]*).*")
+
+def get_args():
+    ap = argparse.ArgumentParser(description="Tool to a the Groovy expression specifying the " +
```
olehermanse: Incomplete sentence, did you mean `add`?

vpodzime: Ooops, fixed.

vpodzime:
> Also, this needs to be backported to 3.7.x and 3.10.x.

And 3.12.x, of course.

PR #385, opened 2018-05-18
ENT-3648 More complex upgrade strategy

Lex-2008:
Two fallback upgrade mechanisms and manual upgrade instructions

Now, how do I test this?

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -305,6 +305,9 @@ if [ ! -d $PREFIX/state/pg/data ]; then
 
   mkdir -p $PREFIX/state/pg/data
   chown -R cfpostgres $PREFIX/state/pg
+  # Note: postgres expects $PWD to be writeable, so all postgres commands
+  # should be executed from cfpostgres-writeable directory.
+  # /tmp is such directory on most cases
   (cd /tmp && su cfpostgres -c "$PREFIX/bin/initdb -D $PREFIX/state/pg/data")
   touch /var/log/postgresql.log
   chown cfpostgres /var/log/postgresql.log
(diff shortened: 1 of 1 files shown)


packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,18 +349,155 @@ if [ ! -d $PREFIX/state/pg/data ]; then
   fi
 fi
 
+check_disk_space() {
+  # checks disk space, prints warning if needed, and returns:
+  # 0 (true) if it's enough disk space
+  # 1 (false) if we need to abort due to low disk space
+  megabytes_free="$(df -PBM $PREFIX | awk 'FNR==2{gsub(/[^0-9]/,"",$4);print $4}')"
+  # note that if df or awk is not installed, or df has different format,
+  # or something else, then $megabytes_free will be empty string
```
nickanderson: It will be an empty string and how does that affect the less than comparison?

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,18 +349,155 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
     cf_console echo "Migration done, cleaning up"
     rm -rf $BACKUP_DIR
   else
-    cf_console echo "Migration failure. Manual interference required."
-    cf_console echo "Old stuff is saved in $BACKUP_DIR. Good luck!"
-    exit 1
-  fi
+    cf_console echo "Migration failed. Backup is saved in $BACKUP_DIR."
+    if [ -f $BACKUP_DIR/db_dump.sql ]; then
+      cf_console echo "Plaintext dump of the database is in $BACKUP_DIR/db_dump.sql file."
```
nickanderson: Lovely thank you for the instructinos here.

nickanderson:
Humm, you could test pgupgrade failure by linking pgupgrade to bin/false

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: Right, probably I should add this to the comment: as it's written today, this function will print a cryptic message and move on like if there is enough space.

This note is a warning for future generations (or future myself) to think about this case when rewriting this function.

Lex-2008:
yes, but not manually - pgupgrade is shipped together with newer version of postgres.

I'll add debug code which reads some external file into a variable and poison all ifs with things like `... or "$test_flag_file_contents" = "fail_at_line_15"`

Lex-2008:
Ok, so testing instructions:
* wait for this build to finish: https://ci.cfengine.com/view/PR%20Pipeline/job/pr-pipeline/575/
* transfer it to testmachine
* in testmachine, install old (pre-12) CFEngine hub and backup it:
```
service cfengine3 stop
cp -a /var/cfengine{,.bak}
```
* run the test:
```
service cfengine3 stop && rm -rf /var/cfengine && cp -a /var/cfengine{.bak,} && service cf-postgres start && dpkg -i cfengine-nova-hub_3.13.0a.*_amd64.deb
```
* repeat the test (above command) with different numbers from 1 to 4 in this file:
```
echo "4" >/var/cfengine.bak/postgres-10-migration-test
```

Explanation (from the commit):
> Here we attempt three upgrade mechanisms, in this order:
> * using pg_upgrade utility
> * running two databases side-by-side (old and new), and dumping from old one into new one via pipe
> * first dumping from old database into *.sql file, and then importing it into new one

And possible values for number in `/var/cfengine.bak/postgres-10-migration-test` file are:
> 0 - do nothing special (usually first method succeeds)
> 1 - fail first method (so we get a chance to run second method)
> 2 - fail first two methods (so we get a chance to run third method)
> 3 - fail first two methods, and third one on "dumping" stage (so we get a chance to see error message with instructions how to *dump* and import manually)
> 4 - fail first two methods, and third one on "importing" stage (so we get a chance to see error message with instructions how to import *existing* dump file)

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,18 +349,184 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
 
+check_disk_space() {
+  # checks disk space, prints warning if needed, and returns:
+  # 0 (true) if it's enough disk space
+  # 1 (false) if we need to abort due to low disk space
+  megabytes_free="$(df -PBM $PREFIX | awk 'FNR==2{gsub(/[^0-9]/,"",$4);print $4}')"
+  # note that if df or awk is not installed, or df has different format,
+  # or something else, then $megabytes_free will be empty string.
+  # Keep this in mind when trying to modify this function!
+  # As of now, in this case we assume that everything is fine and return 0.
```
vpodzime: But I think we should still avoid the invalid arguments for the ``[ ]`` below in such cases.

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,18 +349,184 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
+}
+
 BACKUP_DIR=$PREFIX/backup-before-postgres10-migration
 if is_upgrade && [ -d "$BACKUP_DIR" ]; then
-  cf_console echo "Migrating database..."
-  (cd /tmp && cf_console time su cfpostgres -c "$PREFIX/bin/pg_upgrade --old-bindir=$BACKUP_DIR/bin --new-bindir=$PREFIX/bin --old-datadir=$BACKUP_DIR/data --new-datadir=$PREFIX/state/pg/data --link")
+  # Here we attempt three upgrade mechanisms, in this order:
+  # * using pg_upgrade utility
+  # * running two databases side-by-side (old and new),
+  #   asnd dumping from old one into new one via pipe
```
vpodzime: *and*?

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,18 +349,184 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
+  # * first dumping from old database into *.sql file,
+  #   and then importing it into new one
+  MIGRATED=0 # flag that migration succeded
+  cf_console echo "Migrating database using pg_upgrade utility..."
+  cf_console echo
+  (cd /tmp && cf_console su cfpostgres -c "$PREFIX/bin/pg_upgrade --old-bindir=$BACKUP_DIR/bin --new-bindir=$PREFIX/bin --old-datadir=$BACKUP_DIR/data --new-datadir=$PREFIX/state/pg/data")
```
vpodzime: I really think all these migration methods should be functions. They are long and the ``if-else-if-else...`` structure below is very hard to read because of that.

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,18 +349,184 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
   if [ $? = 0 ]; then
+    MIGRATED=1
+  else
+    cf_console echo "Migration using pg_upgrade failed."
+    cf_console echo
+    if check_disk_space; then
```
vpodzime: Why is the disk space needed for this "piped" method?

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,18 +349,184 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
+        fi
+      )
+      pipe_result=$?
+      if [ $pipe_result = 0 ]; then
+        MIGRATED=1
+      else
+        cf_console echo "Migration using dumpall | psql failed."
+        cf_console echo
+        if check_disk_space; then
+          cf_console echo "Migrating database using dumpall && psql way..."
```
vpodzime: Isn't this better/more reliable method than the previous one if there's enough disk space?

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,18 +349,184 @@ if [ ! -d $PREFIX/state/pg/data ]; then
   fi
 fi
 
+check_disk_space() {
```
vpodzime: I think this function should take the required space as an argument.

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,18 +349,184 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
+          exit 2
+        fi
+      )
+      pipe_result=$?
+      if [ $pipe_result = 0 ]; then
+        MIGRATED=1
+      else
+        cf_console echo "Migration using dumpall | psql failed."
+        cf_console echo
+        if check_disk_space; then
```
vpodzime: We should check that the actually required amount of space is available. We can use ``pg_dumpall|wc -c`` to get the size of the SQL dump.

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -373,11 +373,27 @@ if is_upgrade && [ -d "$BACKUP_DIR" ]; then
...
+  else
+    SKIPS=0
+  fi
+  export SKIPS
   MIGRATED=0 # flag that migration succeded
   cf_console echo "Migrating database using pg_upgrade utility..."
   cf_console echo
   (cd /tmp && cf_console su cfpostgres -c "$PREFIX/bin/pg_upgrade --old-bindir=$BACKUP_DIR/bin --new-bindir=$PREFIX/bin --old-datadir=$BACKUP_DIR/data --new-datadir=$PREFIX/state/pg/data")
-  if [ $? = 0 ]; then
+  if [ $? = 0 -a $SKIPS -lt 1 ]; then
```
vpodzime: This doesn't look to me like it skips anything, it just ignores the results of the attempts, right? Test results for the following methods can thus be affected by changes done by previous methods.

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: Right, sounds good!

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: as we discussed off-line, this is not a check _before_ "piped" method, but a check _after_ pg_upgrade method failed. If pg_upgrade method failed due to lack of disk space, all following methods are doomed to fail, too.

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: Yes, it is, but it requires unknown (or, as you suggested below, `pg_dumpall|wc -c`) amount of disk space more than the pipe way.

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: Agree, but let's create a separate task for this.

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: Good idea, but I wonder how slow `pg_dumpall|wc -c` might be on production system.

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: yes, right, so maybe it should be renamed to $FAILS. But later (with values of 3 and 4) the same variable is used as just index of method that should fail. So maybe it should be just renamed to $DEBUG.

Another open question is if we want to keep this debug code in released version at all.

Regarding "test results for the following methods can be affected by changes done by previous methods" - my hope is that previous methods don't change source (old) database, and as for target (new) database - I `initdb` it anew before each migration attempt.

PR #393, opened 2018-05-30
Use proper gitconfig and gitattributes file paths when building git

vpodzime:
These two ./configure options are:

  --with-gitconfig=VALUE  Use VALUE instead of /etc/gitconfig as the global
                          git configuration file. If VALUE is not fully
                          qualified it will be interpreted as a path relative
                          to the computed prefix at runtime.

  --with-gitattributes=VALUE
                          Use VALUE instead of /etc/gitattributes as the
                          global git attributes file. If VALUE is not fully
                          qualified it will be interpreted as a path relative
                          to the computed prefix at runtime.

so they should be file paths, not directories. New versions of
git are sensitive to this difference.

Ticket: ENT-3818

### DIFF
--- a/deps-packaging/git/cfbuild-git.spec
+++ b/deps-packaging/git/cfbuild-git.spec
@@ -18,7 +18,7 @@ AutoReqProv: no
 mkdir -p %{_builddir}
 %setup -q -n git-%{git_version}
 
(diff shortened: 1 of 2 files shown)


vpodzime:
I just tested that our [build of *git* with these changes](http://52.91.214.185/packages/testing-pr/jenkins-pr-pipeline-623/PACKAGES_HUB_x86_64_linux_redhat_7/) works just fine.

nickanderson:
Woot

PR #397, opened 2018-06-08
Disable parallel builds on AIX

vpodzime:
There seem to be some issues with numbers of processes running.


vpodzime:
Unfortunately, this doesn't make a difference.

PR #399, opened 2018-06-13
Use gmake when building libcurl

vpodzime:
We need to find out how the environment looks like to reproduce
the same conditions just outside of rpmbuild.


vpodzime:
This is not needed.

PR #401, opened 2018-06-22
ENT-3898: Move old shared/GUI files aside during upgrade 

vpodzime:
Otherwise they may end up being copied over to htdocs together
with the new files because the postinstall.sh script is run when
*both* the old package and the new package are installed.


cf-bottom:

PR #406, opened 2018-06-29
ENT-3921: Make runalerts.php executable.

aleksandrychev:
Because cf-runalerts.service executes this script.


cf-bottom:

PR #416, opened 2018-07-30
Disable Solaris 9 builds on master

vpodzime:
Solaris 9 is not a supported system for future releases.


cf-bottom:

PR #427, opened 2018-08-22
lmdb: Do not put functions into a separate text section on PPC64

vpodzime:
Somehow this breaks things horribly and such functions seem to
collide in address space with 'glibc' functions. Segfaults then
happen and backtraces are totally weird.(*)

(*) at least on RHEL 6.9, gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-23)

### DIFF
--- a/deps-packaging/lmdb/0007-Do-not-put-functions-into-a-separate-text-section-on.patch
+++ b/deps-packaging/lmdb/0007-Do-not-put-functions-into-a-separate-text-section-on.patch
@@ -1,0 +1,35 @@
+From 5ba8c511f0e9c0bd636c8af249d3b821c22259e2 Mon Sep 17 00:00:00 2001
+From: Vratislav Podzimek <vratislav.podzimek@northern.tech>
+Date: Wed, 22 Aug 2018 16:47:04 +0200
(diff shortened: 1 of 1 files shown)


cf-bottom:

craigcomstock:
Interesting! Would be fun to know why exactly. How did you arrive at the solution?

vpodzime:
> Interesting! Would be fun to know why exactly. How did you arrive at the solution?

Based on the segfaults ending up in very weird and unrelated places in the process memory (*glibc* functions) after a `e = calloc(1, MDB_env)` call inside one of these environment functions.

deps-packaging/lmdb/0007-Do-not-put-functions-into-a-separate-text-section-on.patch
```diff
@@ -0,0 +1,35 @@
+From 5ba8c511f0e9c0bd636c8af249d3b821c22259e2 Mon Sep 17 00:00:00 2001
+From: Vratislav Podzimek <v.podzimek@mykolab.com>
```
olehermanse: Should use work e-mail here.

PR #428, opened 2018-08-24
ENT-3972: Save and restore state on package upgrade

vpodzime:
We shouldn't blindly start all our daemons/services after package
upgrade, but rather honor the original/pre-upgrade state.

Changelog: Only start previously running processes on package upgrade

### DIFF
--- a/packaging/common/script-templates/script-common.sh
+++ b/packaging/common/script-templates/script-common.sh
@@ -40,3 +40,33 @@ case "$PKG_TYPE" in
     esac
     ;;
 esac
+
+get_cfengine_state() {
+    if type systemctl >/dev/null 2>&1; then
+        systemctl list-units -l | grep -P '^\s*cf-' | sed -r 's/\s*(cf-[-a-z]+)\.service.*/\1/'
+    else
+        platform_service cfengine3 status | grep 'is running' | sed -r 's/^([-a-z]+).*/\1/'
+    fi
+}
+
+restore_cfengine_state() {
+    # $1 -- file where the state to restore is saved (see get_cfengine_state())
+
+    if type systemctl >/dev/null 2>&1; then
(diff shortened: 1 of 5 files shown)


cf-bottom:

packaging/common/script-templates/script-common.sh
```diff
@@ -40,3 +40,33 @@ case "$PKG_TYPE" in
...
+        systemctl list-units -l | grep '^cf-' | sed -r 's/(cf-[-a-z]+)\.service.*/\1/'
+    else
+        platform_service status cfengine3 | grep 'is running' | sed -r 's/^([-a-z]+).*/\1/'
+    fi
+}
+
+restore_cfengine_state() {
+    # $1 -- file where the state to restore is saved (see get_cfengine_state())
+
+    if [ -x /bin/systemctl ]; then
```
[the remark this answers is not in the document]
vpodzime: According to the other scripts we have it is. But I think using `type` or `which` would make more sense so I'll change it here.

vpodzime:
@olehermanse I fixed a few more things, please re-review.

build-scripts/set_github_status.sh
```diff
@@ -0,0 +1,74 @@
...
+#     $5 - URL to link from the status (e.g. $JOB_URL of the jenkins job)
+#   Or:
+#     $1 - where to get repos and PRs info from
+#     $2 - path to a JSON file ready to POST to GH
+# Env:
+#   $GITHUB_STATUS_TOKEN - token for GitHub authentication
+
+PRs_file="$1"
+if [ -z "$PRs_file" ]; then
+    exit 1
```
olehermanse: Maybe add example usage here, or at least in comment above?

vpodzime: This commit somehow sneaked in to the PR, it's on *master* already.

PR #429, opened 2018-08-28
Fix travis errors

Lex-2008:


Lex-2008:
must be cherry-picked, too

PR #434, opened 2018-09-19
[3.12.x] Try harder to stop PostgreSQL on upgrades

vpodzime:
If 'pg_ctl stop -m smart' fails, it means that there are some
clients that are not properly terminating their sessions. Using
'pg_ctl stop -m fast' shuts down such sessions. That may
potentially be unsafe, but the only clients of our PostgreSQL are
our components -- namely Redis, cf-hub and Mission Portal. All of
these should be dead by the time we stop PostgreSQL so it's safe
to terminate any sessions they may have left open.

(cherry picked from commit 0c3182c65c1764aadf878cf881b9906e9100e1fa)


vpodzime:
@nickanderson are you okay with this cherry-pick of #433 ?

vpodzime:
Revert if not. :smile:

PR #435, opened 2018-09-21
Set the flag for the enterprise init script in state restore

vpodzime:
The enterprise init script checks is was called from an expected
place (the core init script or state restore) so we need to tell
it that we are restoring state in scriptlets.


cf-bottom:

vpodzime:
> and don't you need to `export` it?

I think I don't because I'm sourcing the file not executing it.

PR #438, opened 2018-09-24
ENT-3715 Only attempt to generate key during install if not present on windows

nickanderson:
If an existing key is present during pacakge install we should not try
to re-generate a new one. Trying to generate a new key when one already
exists will fail, and that will cause the package to not finish
installing correctly.

Changelog: Title

### DIFF
--- a/packaging/cfengine-nova/cfengine-nova.wxs
+++ b/packaging/cfengine-nova/cfengine-nova.wxs
@@ -35,7 +35,7 @@
 
 	  <!-- Run cf-key after new installation -->
 	  <Property Id='GenerateKey' Value='empty, for now'  />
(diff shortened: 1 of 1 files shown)


cf-bottom:

nickanderson:
@cf-bottom jenkins

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1199/)

https://ci.cfengine.com/job/pr-pipeline/1199/

vpodzime:
Horaaay! Well, as long as it works...

nickanderson:
I was able to install the package by simply double-clicking as administrator without error.

PR #439, opened 2018-09-25
ENT-3715 Only attempt to generate key during install if not present on windows

nickanderson:
If an existing key is present during pacakge install we should not try
to re-generate a new one. Trying to generate a new key when one already
exists will fail, and that will cause the package to not finish
installing correctly.

Changelog: Title
(cherry picked from commit b08245956187e8ca2e361839b43156da0e3e7d80)


nickanderson:
@cf-bottom jenkins

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/1221/)

https://ci.cfengine.com/job/pr-pipeline/1221/

PR #444, opened 2018-09-26
postgresql.conf handling improvements

vpodzime:
A few improvements for installing, upgrading and patching the *postgresql.conf* file. See individual commit messages for details.


nickanderson:
LGTM

vpodzime:
I want to give this a few more rounds of testing, but I believe it is ready for a (re-)review.

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,10 +350,58 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
+        cf_console echo "Please review the $PREFIX/state/pg/data/postgresql.conf file and update it accordingly."
+        cf_console echo "The original file was saved as $PREFIX/state/pg/data/postgresql.conf.old"
+        mv "$PREFIX/state/pg/data/postgresql.conf" "$PREFIX/state/pg/data/postgresql.conf.old"
+        if [ -n "$new_pgconfig_file" ]; then
+          # Use the new recommended config.
+          cp "$new_pgconfig_file" "$PREFIX/state/pg/data/postgresql.conf"
+          chown cfpostgres "$PREFIX/state/pg/data/postgresql.conf"
+        fi
+        # else -- let PostgreSQL use its defaults
+      fi
```
craigcomstock: maybe change this comment to a cf_console log just for verbosity and awareness to the user?

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,10 +350,58 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
     else
-      cf_console echo "Warning: not enough total memory needed to set shared_buffers=2GB"
+      # No user-modified file, just use the recommended config (see the memory
+      # size checks above) or defaults
```
craigcomstock: here again and maybe a few other places above you might convert your comments into cf_console logging for clarity to the user about what is happening and opportunities for us to diagnose failures when this is run and breaks.

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,10 +350,58 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
-      chown cfpostgres $PREFIX/state/pg/data/postgresql.conf
+      new_pgconfig_file="$PREFIX/share/postgresql/postgresql.conf.cfengine"
+      if ! is_upgrade; then
+        # Not an upgrade, just use the recommended file. (see below for upgrade behavior)
+        cp "$new_pgconfig_file" $PREFIX/state/pg/data/postgresql.conf
+        chown cfpostgres $PREFIX/state/pg/data/postgresql.conf
+      fi
+    else
+      cf_console echo "Warning: not enough total memory needed to use the"
+      cf_console echo "recommended PostgreSQL configuration, using the defaults."
```
nickanderson: Perhaps it would be good to identify this as as our recomendation.

...
CFEngine recommended PostgreSQL configuration, using the PostgreSQL defaults.

packaging/common/cfengine-hub/postinstall.sh
vpodzime: I think the effect of `set -x` makes the `/var/log/CFEngineInstall.log` file good enough to see what was going on. I was trying to avoid duplicate messages because the user already gets a warning that the defaults are being used instead of recommendations due to the lack of RAM.

packaging/common/cfengine-hub/postinstall.sh
vpodzime: Done.

vpodzime:
Sorry for bothering you again guys, but I changed the stuff quite significantly. Still the same ideas, the same behavior, just the script is nicer and the logic in it is a lot more easy to see, I believe. Suggestions addressed, waiting for (hopefully) final builds for testing. I also added a commit that makes `pg_upgrade` output to go into the logs instead of the console because it is much more helpful to us than the user and it produced a lot of noise in the upgrade progress/status reporting thus hiding the important things.

vpodzime:
OK, tested all the possible combinations of intact/modified-compatible/modified-incompatible I could come up with. Everything seems to be working fine. So I'm done with this, just waiting for the final waive.

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,10 +337,66 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
+      cf_console echo "Warning: not enough total memory needed to use the CFEngine"
+      cf_console echo "recommended PostgreSQL configuration, using the defaults."
+      echo "$PREFIX/share/postgresql/postgresql.conf.sample"
+    fi
+  fi
+}
+
+#POSTGRES RELATED
+BACKUP_DIR=$PREFIX/backup-before-postgres10-migration
+if [ ! -d $PREFIX/state/pg/data ]; then
```
craigcomstock: is postgres10 accurate always? Maybe capture the version of use some cfengine version instead?

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,10 +337,66 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
+    pgconfig_type="PostgreSQL default"
+  fi
+
+  if ! is_upgrade; then
+    # Not an upgrade, just use the recommended or default file (see generate_new_postgres_conf())
+    cp -a "$new_pgconfig_file" $PREFIX/state/pg/data/postgresql.conf
+    chown cfpostgres $PREFIX/state/pg/data/postgresql.conf
+  else
+    # Determine which postgresql.conf file to use and put it in the right place.
+    if [ -f "$BACKUP_DIR/data/postgresql.conf.modified" ]; then
```
craigcomstock: How can this be an upgrade if the pg_state dir didn't exist? i.e. the if on line 351 ```if [ ! -d $PREFIX/state/pg/data ]; then```

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -349,10 +337,66 @@ if [ ! -d $PREFIX/state/pg/data ]; then
...
+        chown cfpostgres "$PREFIX/state/pg/data/postgresql.conf.new"
+      else
+        # Failed to start, move the old file aside and use the new one.
+        mv "$PREFIX/state/pg/data/postgresql.conf" "$PREFIX/state/pg/data/postgresql.conf.old"
+        cf_console echo "Warning: failed to use the old postgresql.conf file, using the $pgconfig_type one."
+        cf_console echo "Please review the $PREFIX/state/pg/data/postgresql.conf file and update it accordingly."
+        cf_console echo "The original file was saved as $PREFIX/state/pg/data/postgresql.conf.old"
+        cp -a "$new_pgconfig_file" "$PREFIX/state/pg/data/postgresql.conf"
+        chown cfpostgres "$PREFIX/state/pg/data/postgresql.conf"
+      fi
```
craigcomstock: I like all the above verbose and informative details! :+1:

packaging/common/cfengine-hub/postinstall.sh
vpodzime: No idea, you'd have to ask @Lex-2008 about it. I'm not trying to fix everything here. ;)

packaging/common/cfengine-hub/postinstall.sh
vpodzime: No idea either. But I think maybe some time ago we used a different directory? Or some other DBMS? Again, not trying to fix everything here. :)

PR #445, opened 2018-09-27
Give httpd time to stop listening on 80/443 on upgrades

vpodzime:
If 'apachectl stop' finishes with exit code 0, but something is
still listening on the HTTP(S) port(s), it's worth it to try and
give 'httpd' some more time to finish terminating properly before
just giving up.

### DIFF
--- a/packaging/common/cfengine-hub/preinstall.sh
+++ b/packaging/common/cfengine-hub/preinstall.sh
@@ -116,8 +116,13 @@ then
     HTTPD_RUNNING=`filter_netstat_listen ":80\s|:443\s"`
     if [ ! -z "$HTTPD_RUNNING" ];
     then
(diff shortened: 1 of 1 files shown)


cf-bottom:

vpodzime:
> would a killall of some sort work?

Well, if I read the scriptlet correctly, we are trying to avoid killing non-CFEngine processes. Note that the sleep I added only happens after `apachectl stop` finishes and only when it exits with the code `0`. Not sure if we should add some more aggressive command to kill the process that we identify as *the CFEngine httpd*.

vpodzime:
I think this should be cherry-picked to *3.12.x*, opinions?

olehermanse:
@vpodzime agreed

vpodzime:
> @vpodzime agreed

#448

PR #448, opened 2018-10-03
[3.12.x] Give httpd time to stop listening on 80/443 on upgrades

vpodzime:
If 'apachectl stop' finishes with exit code 0, but something is
still listening on the HTTP(S) port(s), it's worth it to try and
give 'httpd' some more time to finish terminating properly before
just giving up.

(cherry picked from commit c0af01d63081f9c529d1c981b2b8a8b2cf0c90cd)


cf-bottom:

PR #450, opened 2018-10-11
Attempt to fix access to tests/acceptance

olehermanse:


cf-bottom:

olehermanse:
@Lex-2008 @vpodzime @nickanderson @craigcomstock This change didn't make a difference. Do you think we should leave it there or revert it?

vpodzime:
> This change didn't make a difference. Do you think we should leave it there or revert it?

Leave it.

PR #455, opened 2018-10-15
Upgrade: assume PostgreSQL is running on 3.7.x if running cf-hub

vpodzime:
3.7.x has broken detection of the PostgreSQL status and
'postgres' is always reported as 'not running'. Thus when we
capture the pre-upgrade state of runnig processes, PostgreSQL is
missing there and as a consequence it is not started when the
upgrade is finished.

To workaround this, let's assume that PostgreSQL is running on
3.7.x if cf-hub is also running.

Ticket: ENT-4082


cf-bottom:

vpodzime:
I'll check this in the upgrade tests and backport it to 3.12.x if it works as expected.

PR #461, opened 2018-10-22
Try to avoid some race conditions when checking if PostgreSQL is running

vpodzime:


cf-bottom:

packaging/common/cfengine-hub/preinstall.sh
```diff
@@ -145,51 +145,66 @@ then
...
-if [ ! -z "$PSQL_RUNNING" ];
-then
-  cf_console echo "There seems to be a server listening on port 5432"
-  cf_console echo "This might mean that there is a PostgreSQL server running on the machine already"
-  cf_console echo "Checking if the Postgres installation belongs to a previous CFEngine deployment"
-  PSQL_COMMAND=$(ps -p $(fuser -n tcp 5432 2>/dev/null) -o args=|cut -d' ' -f1)
-  if [ ! -z "$PSQL_COMMAND" ];
+ensure_postgres_terminated() {
+  PSQL_RUNNING=`filter_netstat_listen ":5432\s"`
+  if [ ! -z "$PSQL_RUNNING" ];
```
Lex-2008: you could also invert this condition and write something like
```suggestion
  if [ -z "$PSQL_RUNNING" ];
  then
    exit 0
  fi
```
and remove one level of indentation from the following commands.

Less indentation - less disk usage - greener environment! :leaves:

packaging/common/cfengine-hub/preinstall.sh
```diff
@@ -145,51 +145,66 @@ then
...
   then
-    if [ "$PSQL_COMMAND" = "$PREFIX/bin/postgres" ];
+    cf_console echo "There seems to be a server listening on port 5432"
+    cf_console echo "This might mean that there is a PostgreSQL server running on the machine already"
+    cf_console echo "Checking if the Postgres installation belongs to a previous CFEngine deployment"
+
+    # First let's try giving PostgreSQL time to terminate (it may still be in
```
Lex-2008: I would expect to see some `sleep 1` when we talk about "giving some time" :timer_clock:

packaging/common/cfengine-hub/preinstall.sh
```diff
@@ -145,51 +145,66 @@ then
...
+    # the process of doing so).
+    pgpid=$(echo "$PSQL_RUNNING" | sed -r -e '/pid=/!d' -e 's/.*pid=([0-9]+),.*/\1/' | tail -1)
+    pgargs=$(ps -p $pgpid -o args=)
+    if [ $? != 0 ];
+    then
+      cf_console echo "The PostgreSQL server terminated, moving on."
+      return 0
+    fi
+    PSQL_COMMAND=$(echo "$pgargs" | cut -d' ' -f1)
+    if [ ! -z "$PSQL_COMMAND" ];
```
Lex-2008: Same here:
```suggestion
    if [ -z "$PSQL_COMMAND" ];
    then
      cf_console echo "There is an unknown process listening on the PostgreSQL port"
      exit 1
    fi
```

packaging/common/cfengine-hub/preinstall.sh
vpodzime: Good point. Changing.

packaging/common/cfengine-hub/preinstall.sh
vpodzime: Oh, that's just a leftover. Good catch.

packaging/common/cfengine-hub/preinstall.sh
vpodzime: This one has a bigger `else` branch, I think the more likely one (which is `! -z "$PSQL_COMMAND"`) should be first. So keeping this one as is.

vpodzime:
> What about stopping PostgreSQL **before** moving its data dir to backup dir?

Yeap, I was thinking that too. But I want to submit is as a separate PR. This one seems to be the last missing piece to make the upgrade deployment tests green.

PR #464, opened 2018-10-24
[3.12.x] upgrade fixes

vpodzime:
Three things that should fix issues with upgrading from older versions.


cf-bottom:

PR #465, opened 2018-10-24
[3.10.x] upgrade fixes

vpodzime:

PR #468, opened 2018-10-30
Get running services from lines matching exactly what we expect

vpodzime:
When saving state we are extracting names of CFEngine services
that are running. The recognized sevices are the ones starting
with the 'cf-' prefix and the 'cfengine3.service'. Everything
else has to be ignored, otherwise the 's//' sed command fails and
we get full lines from the 'systemct list-units' output in the
state snapshot which is supposed to only list the running
services and nothing else.

Ticket: ENT-4082

### DIFF
--- a/packaging/common/script-templates/script-common.sh
+++ b/packaging/common/script-templates/script-common.sh
@@ -43,7 +43,7 @@ esac
 
 get_cfengine_state() {
     if type systemctl >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


cf-bottom:

vpodzime:
Added to commit to still-to-be-merged PRs https://github.com/cfengine/buildscripts/pull/464 and https://github.com/cfengine/buildscripts/pull/465

PR #477, opened 2018-11-09
ENT-4231: Preserve the 'pg_hba.conf' file on upgrades

vpodzime:
This file defines access control for PostgreSQL and must be
preserved on upgrade. It may be granting extra access (e.g. in
the HA setup), but it may also be blocking some of the defaults.

Changelog: None


cf-bottom:

PR #485, opened 2018-11-19
[3.12.x] CFE-2949: Do not use '-r' with 'sed' when saving state on upgrade

vpodzime:
Some platforms don't have this fancy feature of their 'sed'
tool. It's okay to use it on the systemd-based platforms though,
these are new enough.

(cherry picked from commit ad3c50df7dcb02724af4a4746c24e27046a20858)


cf-bottom:

PR #488, opened 2018-11-19
[3.12.x] CFE-2949: Use 'awk' instead of 'sed' to save state on upgrade

vpodzime:
'sed' doesn't support the '-r' option on old platforms (exotics)
and also its interpretation of regular expressions is a
mystery. 'awk' seems to work much more reliably.

(cherry picked from commit 9ce8e6dfcb5a1d37ed5ceca81d244cdafffcc15a)


cf-bottom:

PR #494, opened 2018-11-21
Refactor how we check if PostgreSQL should be prepared for migration

vpodzime:
Having the same check in two places is error prone to not
updating both places with changes. Exactly as it happened
recently in 7c515c832b.

Doing the check once and storing the result is better and will
also make it easier to change the check to something more complex
doing decisions on version differences and not hardcoded values.

### DIFF
--- a/packaging/common/cfengine-hub/preinstall.sh
+++ b/packaging/common/cfengine-hub/preinstall.sh
@@ -15,10 +15,16 @@ if is_upgrade; then
     fi
 fi
 
-BACKUP_DIR=$PREFIX/backup-before-postgres11-migration
-
 # If upgrading from a version below 3.13 that has PostgreSQL, and the data dir exists.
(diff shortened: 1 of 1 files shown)


cf-bottom:

vpodzime:
Works!

packaging/common/cfengine-hub/preinstall.sh
```diff
@@ -15,10 +15,16 @@ if is_upgrade; then
     fi
 fi
 
-BACKUP_DIR=$PREFIX/backup-before-postgres11-migration
-
 # If upgrading from a version below 3.13 that has PostgreSQL, and the data dir exists.
 if is_upgrade && egrep '^3\.([6-9]|1[012])\.' "$PREFIX/UPGRADED_FROM.txt" >/dev/null && [ -d "$PREFIX/state/pg/data" ]; then
+  alias migrating_postgres='true'
+else
+  alias migrating_postgres='false'
```
olehermanse: 🤦‍♂️

PR #496, opened 2018-11-29
Check for presence of $BACKUP_DIR/data, not $BACKUP_DIR, when restoring DB

Lex-2008:


cf-bottom:

PR #498, opened 2018-11-29
Fix platform_service() on AIX in RPM packages

vpodzime:
The 'cfengine3' init script is under /etc/rc.d/init.d on AIX. We
already have functions for these tweaks, let's just use them
more.

Ticket: ENT-4305
Changelog: Fix running cfengine3 init script path in AIX RPMs


cf-bottom:

PR #500, opened 2018-11-30
WIP: Close the console FD on upgrade

vpodzime:


cf-bottom (changes requested):
Please use a company e-mail instead of v***@mykolab.com

PR #501, opened 2018-12-03
ENT-2614 Clean php extensions directory before install

Lex-2008:


cf-bottom:

packaging/common/cfengine-hub/preinstall.sh
```diff
@@ -347,6 +347,10 @@ if [ -d $PREFIX/httpd/htdocs ]; then
   find $PREFIX/httpd/htdocs -depth -type d -exec rmdir {} \;
 fi
 
+if [ -d $PREFIX/httpd/php/lib/php/extensions/no-debug-non-zts-20170718 ]; then
+  rm $PREFIX/httpd/php/lib/php/extensions/no-debug-non-zts-20170718/*
+fi
```
vpodzime: That's a very weird directory name. Any idea where it comes from?

Lex-2008: Not really. The only mention of it that I found was in https://github.com/cfengine/buildscripts/blob/master/packaging/common/cfengine-hub/postinstall.sh#L70

craigcomstock: I believe it is constructed from the bits of info about the extension:

- no-debug
- non-zts (zts is Zen Thread Safety)
- date stamp

https://serverfault.com/questions/399310/php-extensions-no-debug-non-zts-xxxxxxxx

vpodzime: OK, thanks for the info @craigcomstock!

PR #502, opened 2018-12-04
Start Redis after upgrade if it was running before it

vpodzime:
Ticket: ENT-4313


vpodzime:
So the problem here is that the `start_redis` function from the enterprise init script requires things from the *core* init script. We thus need to source the *core* init script in the state restoration procedures.

Lex-2008:
I the deployment test failures I saw that `cf-consumer` also failed to start. Or was it because it depends on redis?

vpodzime:
> But let me double-check it is started with these changes.

Yup, it is started just fine.

PR #504, opened 2018-12-11
ENT-4126: Set phar read only off for PHP

aleksandrychev:
New Export/Import tooling requires phar

### DIFF
--- a/deps-packaging/php/debian/rules
+++ b/deps-packaging/php/debian/rules
@@ -139,6 +139,9 @@ install: build
 	# Set the default timezone for php to UTC
 	sed -ri 's/^(\s|;)*date.timezone\s*=.*/date.timezone = "UTC"/g' $(CURDIR)/debian/tmp$(PREFIX)/httpd/php/lib/php.ini
 
+	# Set the phar readonly Off for php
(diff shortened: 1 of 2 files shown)


cf-bottom:

deps-packaging/php/debian/rules
```diff
@@ -139,6 +139,9 @@ install: build
 	# Set the default timezone for php to UTC
 	sed -ri 's/^(\s|;)*date.timezone\s*=.*/date.timezone = "UTC"/g' $(CURDIR)/debian/tmp$(PREFIX)/httpd/php/lib/php.ini
 
+	# Set the phar readonly Off for php
+    sed -ri 's/^(\s|;)phar.readonly = On/phar.readonly = Off/g' $(CURDIR)/debian/tmp$(PREFIX)/httpd/php/lib/php.ini
```
Lex-2008: this file is makefile, so it must be indented with **tab**, not _spaces_

aleksandrychev: ok, thanks!

craigcomstock:
you don't need trailing `/g` right? there's only one instance of the search pattern.

aleksandrychev:
>you don't need trailing /g right? there's only one instance of the search pattern.

@craigcomstock yes, right

PR #510, opened 2019-01-04
Set -e in package scriptlets

vpodzime:
All the failures in the scriptlets should generally be fatal. We
may need to identify some exceptions where we will do 'set +x'
locally.

Ticket: ENT-3385


cf-bottom:

vpodzime:
There's some scriptlet failure on at least HP-UX.

PR #513, opened 2019-01-21
Give PostgreSQL processes extra time to terminate

vpodzime:
Some older versions of pg_ctl seem to leave a couple postgres
processes running when 'pg_ctl stop' finishes. In a second, they
are all gone. But if we are too quick checking if there is
something listening on the port 5432, we may stop upgrade process
for no good reason.

Changelog: None


cf-bottom:

vpodzime:
Unfortunately, this doesn't seem to help. :disappointed:

vpodzime:
There's a different problem here and not on *master*.

PR #515, opened 2019-02-04
INVALID: Let PostgreSQL pick up system libreadline

vpodzime:
Maybe it just works?

### DIFF
--- a/deps-packaging/postgresql-hub/cfbuild-postgresql-hub.spec
+++ b/deps-packaging/postgresql-hub/cfbuild-postgresql-hub.spec
@@ -23,7 +23,8 @@ mkdir -p %{_builddir}
 
 SYS=`uname -s`
 
-LD_LIBRARY_PATH=%{prefix}/lib CPPFLAGS=-I%{prefix}/include ./configure --prefix=%{prefix} --without-zlib --without-readline --with-openssl
+# let PostgreSQL pick up system libreadline
+LD_LIBRARY_PATH=%{prefix}/lib CPPFLAGS=-I%{prefix}/include ./configure --prefix=%{prefix} --without-zlib --with-openssl
 
 if [ -z $MAKE ]; then
   MAKE_PATH=`which make`
--- a/deps-packaging/postgresql-hub/debian/rules
+++ b/deps-packaging/postgresql-hub/debian/rules
@@ -12,7 +12,7 @@ build: build-stamp
 build-stamp:
 	dh_testdir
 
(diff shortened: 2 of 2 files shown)


cf-bottom:

vpodzime:
> Maybe it just works?

No, it doesn't.

BasementTrix:
On Mon, Feb 04, 2019 at 07:23:41AM -0800, Vratislav Podzimek wrote:
>   * INVALID: Let PostgreSQL pick up system libreadline

But why?  AFAIK, libreadline is only used in psql -- and the individual
can use their own client -- which may not even be on the policy host.

-- 
John D. "Trix" Farrar            __\\|//__               Basement.NET
trix@basement.net                (` o-o ')   http://www.basement.net/
--------------------------------ooO-(_)-Ooo--------------------------
 GPG Key Fprint: 39D0 BCD2 9C9F B7EF B125  D419 61E8 8648 16F2 1814

vpodzime:
> But why? AFAIK, libreadline is only used in psql

Yes, that's exactly where I (and some other people) want to use it. :)

BasementTrix:
On Tue, Feb 05, 2019 at 08:37:04AM -0800, Vratislav Podzimek wrote:
> Date: Tue, 05 Feb 2019 08:37:04 -0800
> > Date: Tues, 05 Feb 2019 trix@basement.net wrote:
> > But why? AFAIK, libreadline is only used in psql
> 
> Yes, that's exactly where I (and some other people) want to use it. :)

To finish the quote ... and the individual can use their own client --
which may not even be on the policy host.

Why not install the PostgreSQL client from OS vendor repositories and
move on with life?  I thought the whole point of the build system is
that it not "just pick up" things from the system.

To the greatest extent possibly, if it's a dependency, the build
system should include it.  The PostgreSQL server builds fine without
readline and the client packages from most OS vendors include it.

At least that's the principle I was told to work with when I was first
working on buildscripts for FreeBSD.

-- 
John D. "Trix" Farrar            __\\|//__               Basement.NET
trix@basement.net                (` o-o ')   http://www.basement.net/
--------------------------------ooO-(_)-Ooo--------------------------
 GPG Key Fprint: 39D0 BCD2 9C9F B7EF B125  D419 61E8 8648 16F2 1814

vpodzime:
You are right, @BasementTrix, sorry for the confusion. This was just for some R&D builds used for experimenting with PostgreSQL and CFEngine. That's why it has said **INVALID** from the very beginning. I was just taking the chances to get an easier to use *psql* with very little efforts.

But this should be done properly at some point. People are using the *psql* utility shipped with CFEngine and it's hard to use without readline support.

PR #520, opened 2019-02-13
Disable pr-pipeline builds on Solaris 10 SPARC64

vpodzime:
The Solaris 10 zone is heavily unreliable and is causing builds
to fail almost every day. It's better to have green builds and
only make sure everything is fine on Solaris 10 before the
release. Note that we are still running tests on Solaris 11
SPARC64.

(cherry picked from commit 7dcfa5190b2d3a8875d1e564747e5e6283303b24)


cf-bottom:

PR #523, opened 2019-02-28
Add an option to get ONLY exotic labels with get_labels_expr.py

vpodzime:
This will allow us to split the full run of tests into two jobs
-- one for non-exotic platforms and one for exotic platforms. We
can then treat their results, stability issues,... differently.

### DIFF
--- a/build-scripts/get_labels_expr.py
+++ b/build-scripts/get_labels_expr.py
@@ -18,6 +18,8 @@ def get_args():
     ap.add_argument("-e", "--run-on-exotics", nargs="?", default=False, const=True,
                     help="Run on exotic labels if RUN_ON_EXOTICS is not specified " +
                     "or it is not one of '0', 'no'.")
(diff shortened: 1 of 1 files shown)


cf-bottom:

build-scripts/get_labels_expr.py
```diff
@@ -31,14 +33,21 @@ def non_empty_lines(file_obj):
         if content:
             yield content
 
-def main(labels_f_path, exotics_f_path, run_on_exotics):
-    labels_to_run = set()
-    with open(labels_f_path, "r") as f:
+def get_labels_from_file(f_path):
+    with open(f_path, "r") as f:
         for label in non_empty_lines(f):
+            yield label
```
Lex-2008: nice!

vpodzime:
I'm merging this because I need it to be able to change the configuration of the `pr-pipeline` job in jenkins.

PR #524, opened 2019-02-28
Add an option to get ONLY exotic labels with get_labels_expr.py

vpodzime:
This will allow us to split the full run of tests into two jobs
-- one for non-exotic platforms and one for exotic platforms. We
can then treat their results, stability issues,... differently.

(cherry picked from commit 31e84179b748baeeb07b01b2adcb4c14442d0197)


vpodzime:
> I'm merging this because I need it to be able to change the configuration of the `pr-pipeline` job in jenkins. (#523)

Same here.

PR #526, opened 2019-03-01
Fix getting configuration filter for exotics only

vpodzime:
We need to respect the original list of labels not just take the
exotics. There might be some exotics that are not in the full
list of labels which thus should be filtered out.

Also make the logic clearer by using sets and set operations.


Lex-2008:
good point!

PR #532, opened 2019-04-08
Drop 'Hub' from the installation log filename

vpodzime:
The log is produced on both hub and client installations. Instead
of adding complexity to differentiate between the cases, let's
just use a neutral filename.

Ticket: ENT-4564

### DIFF
--- a/packaging/common/script-templates/script-common-header-last.sh
+++ b/packaging/common/script-templates/script-common-header-last.sh
@@ -21,7 +21,7 @@ is_nova()
   test "$PROJECT_TYPE" = "cfengine-nova" || test "$PROJECT_TYPE" = "cfengine-nova-hub"
 }
 
-INSTLOG=/var/log/CFEngineHub-Install.log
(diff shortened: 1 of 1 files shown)


vpodzime:
@nickanderson, how much do we have to care about places where this might be hardcoded? (that are out of our control).

olehermanse:
> @nickanderson, how much do we have to care about places where this might be hardcoded? (that are out of our control).

@vpodzime I would care enough to make this 3.14+ only

vpodzime:
See also: https://github.com/cfengine/masterfiles/pull/1362

nickanderson:
> @nickanderson, how much do we have to care about places where this might be hardcoded? (that are out of our control).

Humm, well we need to be sure to update docs. Probably, it's a large enough change that we should do it for the next minor release only, so just in master.

vpodzime:
> Let's be sure to merge with updated docs as well.

Sir! Yes sir! https://github.com/cfengine/documentation/pull/2102

PR #533, opened 2019-04-09
DONOTMERGE: do not strip binaries

vpodzime:


cf-bottom:

cf-bottom (changes requested):
Please use a company e-mail instead of v***@mykolab.com

PR #549, opened 2019-05-08
State directory permissions fixes

vpodzime:


vpodzime:
> guess we can't mention `cfpostgres` user in spec file?

Yeah, I'm afraid it's not created by that time.

PR #558, opened 2019-05-20
Add some flags for building Perl on AIX 7

vpodzime:
Otherwise it fails to load the POSIX.pm module which loads the
POSIX.so shared object and fails to find some standard C
functions.

Ticket: ENT-4659
Changelog: None


cf-bottom:

vpodzime:
The [console output](https://ci.cfengine.com/job/testing-pr/label=PACKAGES_ppc64_aix_71/3666/consoleFull) from the build which (re)built both Perl and OpenSSL proves this works.

PR #568, opened 2019-06-27
Add a patch fixing x86_64 assembly in OpenSSL

vpodzime:
Upstream has this fix, but it didn't make it to OpenSSL-1.1.1b so
we have to apply it ourselves.

(cherry picked from commit 3c0155f0a50c323199a079f6d77b13da6bcc9a7c)


Lex-2008:
ACK!

PR #569, opened 2019-06-27
ENT-4821: Fixed OOTB https redirection when hostname -f contains uppercase

nickanderson:
The output of `hostname -f` was used as part of the ssl certificate filename
while the mustache template rendered by cf-agent uses sys.fqhost which is always
lowercase. This mis-alignment resulted in the rendered httpd.conf not
validating, therefore not updating the live configuration to redirect all http
traffic to https.

Changelog: Title
Ticket: ENT-4821


nickanderson:
merge or wait until after 3.14?

vpodzime:
> merge or wait until after 3.14?

Merge, but it will probably not make it in anyway.

PR #570, opened 2019-07-01
ENT-4288 Update dependency tables after May-July 2019 releases

Lex-2008:


cf-bottom:

PR #574, opened 2019-07-12
ENT-4853 Wait for cf-postgresql service to be running before rotating keys

Lex-2008:


cf-bottom:

Lex-2008:
@cf-bottom build please

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -985,7 +985,7 @@ else
     cf_console platform_service cfengine3 start
 fi
 
-if is_upgrade; then
+if is_upgrade && wait_for_cf_postgres; then
     true "Rotating keys..."
```
craigcomstock: Should it exit if postgres is not available? Right now it just moves on if postgres isn't available for 5 seconds right?

Lex-2008: `wait_for_cf_postgres` returns 1 if postgres is not available after waiting for 5 minutes, so expression inside `if` condition evaluates to false, and rotating keys doesn't happen.

That's how it should be, and I think that's how it happens. Or does it look differently from the code? :-)

On the other side I now think that maybe `wait_for_cf_postgres` should try connecting to postgresql instead of checking its status - in case someone uses remote database with help of port and socket forwarding.

Lex-2008:
build: https://ci.cfengine.com/job/pr-pipeline/2349/

Lex-2008:
rebuild: https://ci.cfengine.com/job/pr-pipeline/2351/

packaging/common/script-templates/script-common.sh
```diff
@@ -81,6 +81,19 @@ restore_cfengine_state() {
     fi
 }
 
+wait_for_cf_postgres() {
+    # wait for CFEngine Postgresql service to be available, up to 5 sec.
+    # Returns 0 is psql command succeeds,
+    # Returns 1 if exited by timeout (service not running after 5 seconds)
+    for 1 2 3 4 5; do
```
vpodzime: `$(seq 1 5)` would be easier to change in the future if we decide we want to wait for 10 seconds.

packaging/common/script-templates/script-common.sh
```diff
@@ -81,6 +81,19 @@ restore_cfengine_state() {
...
+        if $PREFIX/bin/psql cfsettings -c "SELECT 1;" >/dev/null 2>&1; then
+            return 0
+        fi
+        sleep 1
+    done
+    return 1
```
vpodzime: I think it might also be worth it to print/log something here.

Lex-2008: agreed, good idea

Lex-2008:
new build: https://ci.cfengine.com/job/pr-pipeline/2359/

Lex-2008:
typo fix build: https://ci.cfengine.com/job/pr-pipeline/2365/

Lex-2008:
Last build succeeds in my test:
```
+ wait_for_cf_postgres
+ seq 1 5
+ true checking if Postgresql is available...
+ /var/cfengine/bin/psql cfsettings -c SELECT 1;
+ true waiting 1 sec for Postgresql to become available...
+ sleep 1
+ true checking if Postgresql is available...
+ /var/cfengine/bin/psql cfsettings -c SELECT 1;
+ true it is, indeed!
+ return 0
+ true Rotating keys...
```

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2369/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2369/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2369/

packaging/common/script-templates/script-common.sh
```diff
@@ -81,6 +81,22 @@ restore_cfengine_state() {
...
+    for i in $(seq 1 5); do
+        true "checking if Postgresql is available..."
+        if $PREFIX/bin/psql cfsettings -c "SELECT 1;" >/dev/null 2>&1; then
+            true "Postgresql is available, moving on"
+            return 0
+        fi
+        true "waiting 1 sec for Postgresql to become available..."
+        sleep 1
+    done
+    return 1
```
vpodzime: Hmmm, this basically waits one extra second in the end. Maybe it should just do `$PREFIX/bin/psql cfsettings -c "SELECT 1;" >/dev/null 2>&1;` instead of the final `return 1`? It will be the last attempt and also set the return code for the function.

Lex-2008: good idea!

PR #584, opened 2019-07-24
ENT-4670: Added schema tests to nova build.

craigcomstock:
merge with https://github.com/cfengine/nova/pull/1465

### DIFF
--- a/build-scripts/travis
+++ b/build-scripts/travis
@@ -45,6 +45,16 @@ sudo apt-get install s3cmd
   echo y
 ) | s3cmd --configure >/dev/null
 
+install_package() {
+    sudo dpkg -i cfe.deb || (
+        exit_code=$?
+        echo "Problem installing CFEngine package $exit_code"
+        echo "=== BEGIN /var/log/CFEngine*-Install.log ==="
+        sudo cat /var/log/CFEngine*-Install.log
+        echo "=== END /var/log/CFEngine*-Install.log ==="
+        return $exit_code
+    )
+}
 
 case "$JOB" in
     ( build )
@@ -128,7 +138,7 @@ case "$JOB" in
 
     ( acceptance-test )
         s3cmd get --no-progress $DEB_PACKAGE cfe.deb
-        sudo dpkg -i cfe.deb
+        install_package
         touch /home/travis/rpmvercmp
         chmod a+x ~/rpmvercmp
         cd ..
@@ -195,7 +205,7 @@ case "$JOB" in
     ( reporting-test )
         sudo apt-get -y install python-psycopg2
         s3cmd get --no-progress $DEB_PACKAGE cfe.deb
-        sudo dpkg -i cfe.deb
+        install_package
         sudo service cfengine3 restart
         ps aux | grep cf-
         s3cmd get --no-progress $TESTD cf-testd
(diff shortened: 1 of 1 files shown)


build-scripts/travis
```diff
@@ -209,6 +209,39 @@ case "$JOB" in
         sudo prove -v *.test
     ;;
 
+    ( schema-test )
+        # NOTE: these tests will manipulate the system "cfdb" database
```
craigcomstock: I need to clean up this file. Am doing now. :p

craigcomstock: cleaned up. will see if travis is still green. :)

build-scripts/travis
```diff
@@ -209,6 +209,23 @@ case "$JOB" in
...
+        # so parallel execution is not possible
+        s3cmd get --no-progress $DEB_PACKAGE cfe.deb
+        sudo dpkg -i cfe.deb || (
+            echo "Problem installing CFEngine package $?"
+            # TODO CFEngineHub-Install.log or CFEngine-Install.log?
+            echo "=== BEGIN /var/log/CFEngine*-Install.log ==="
+            sudo cat /var/log/CFEngine*-Install.log
+            echo "=== END /var/log/CFEngine*-Install.log ==="
+            ) # TODO bail out early?
```
vpodzime: Yes, why should it continue? AFAICT, it should just fail.

Lex-2008: I agree. if CFEngine installs, but `dpkg` returs non-0 status - it's better to be aware of this. And green Travis build is the best way to become **un**aware :-)

craigcomstock: Agreed. I think I had put an exit in there and now realize it wasn't working because I was exiting the sub-shell. Will fix! :)

Lex-2008: good point. Maybe worth wrapping it into "proper" `if … then … fi` instead of "simple" `|| ( … )`?

craigcomstock: I refactored to a shared `install_package` function and am testing this out. No reason to let the other JOB types fail silently.

craigcomstock:
> Make sure the test fails if dpkg fails, as mentioned by @vpodzime.

Just broke it with a bad filename. That works. Now will go back to just broken schema.sql which should (sadly) allow cfe.deb to install just fine (but be broken).

```
+install_package
+sudo dpkg -i cfe.deb.broken
dpkg: error processing archive cfe.deb.broken (--install):
 cannot access archive: No such file or directory
Errors were encountered while processing:
 cfe.deb.broken
+exit_code=1
+echo 'Problem installing CFEngine package 1'
Problem installing CFEngine package 1
+echo '=== BEGIN /var/log/CFEngine*-Install.log ==='
=== BEGIN /var/log/CFEngine*-Install.log ===
+sudo cat '/var/log/CFEngine*-Install.log'
cat: /var/log/CFEngine*-Install.log: No such file or directory
The command "buildscripts/build-scripts/travis" exited with 1.
Done. Your build exited with 1.
```

craigcomstock:
Build also fails if schema.sql is broken.

```
+ /var/cfengine/httpd/php/bin/php /var/cfengine/httpd/htdocs/index.php cli_tasks inventory_refresh
Error:: A Database Error Occurred
Error Number: 42P01/7,ERROR:  relation "v_inventory" does not exist
LINE 1: INSERT INTO __inventory (SELECT * FROM v_inventory WHERE val...
                                               ^
QUERY:  INSERT INTO __inventory (SELECT * FROM v_inventory WHERE values IS NOT NULL)
CONTEXT:  PL/pgSQL function update_inventory() line 12 at SQL statement,SELECT * FROM update_inventory(),Filename: core/Cf_DB_pdo_driver.php,Line Number: 93+echo '=== END /var/log/CFEngine*-Install.log ==='
=== END /var/log/CFEngine*-Install.log ===
+return 1
The command "buildscripts/build-scripts/travis" exited with 1.
```

Notice that if dpkg -i fails the install log is printed out between `=== BEGIN` and `=== END`.

craigcomstock:
@vpodzime @olehermanse RTM (ready-to-merge) as it were. :+1: :bowing_man:

build-scripts/travis
```diff
@@ -45,6 +45,16 @@ sudo apt-get install s3cmd
...
 ) | s3cmd --configure >/dev/null
 
+install_package() {
+    sudo dpkg -i cfe.deb || (
+        exit_code=$?
+        echo "Problem installing CFEngine package $exit_code"
+        echo "=== BEGIN /var/log/CFEngine*-Install.log ==="
+        sudo cat /var/log/CFEngine*-Install.log
+        echo "=== END /var/log/CFEngine*-Install.log ==="
+        return $exit_code
```
vpodzime: I think this `return` only returns from the subshell. But it doesn't matter too much because that sets the exit/return code of the function anyway. To eliminate the confusion I think you could make this just a block `{}` instead of a subshell.

PR #586, opened 2019-07-29
ENT-4758 ￼Create tarball with a filesystem image and accompanying files

Lex-2008:


cf-bottom:

Lex-2008:
@cf-bottom build it on RHEL, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2414/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2414/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2414/

Lex-2008:
rebuild: https://ci.cfengine.com/job/pr-pipeline/2421/

Lex-2008:
TODO: mv in install.py should preserve file sparseness

packaging/cfengine-nova/coreos/install.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+if [[ $EUID -ne 0 ]]; then
+  echo "This script must be run as root"
+  exit 1
+fi
+
+cd "$(dirname "$0")"
+
+# copy image to expected logation
+cp cfengine3.img /var
```
craigcomstock: s/logation/location/

Lex-2008:
good build: https://ci.cfengine.com/job/pr-pipeline/2447/
the package to test: http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2447/PACKAGES_x86_64_linux_redhat_7/cfengine-nova-3.15.0a.a57920bcd.x86_64.fs-img.pkg.tar.gz

build-scripts/package
```diff
@@ -156,6 +156,29 @@ case "$PACKAGING" in
         mv usr etc var/cfengine/share/
         tar czvf $TARBALL .  > $TARBALL.filelist
     fi
+    if [ -d $P/coreos ]
```
craigcomstock: running shellcheck on this script it warned about line 155 above `[ -d $PKG-* ]` and when I try that I get an error. I wonder if we create a TAR package do we get the same error?

```
craig@other:~/tmp$ mkdir foo-1
craig@other:~/tmp$ mkdir foo-2
craig@other:~/tmp$ mkdir foo-3
craig@other:~/tmp$ PKG="foo"
craig@other:~/tmp$ [ -d $PKG-* ]
-bash: [: too many arguments
```

or rather with sh:
```
$ cd tmp
$ PKG="foo"
$ [ -d $PKG-* ]
sh: 5: [: foo-1: unexpected operator
```

build-scripts/package
```diff
@@ -156,6 +156,29 @@ case "$PACKAGING" in
...
+    then
+        # Create filesystem image
+        IMAGE=$P/coreos/cfengine3.img
```
craigcomstock: My first thought was shellcheck complaining to "Double quote to prevent globbing and word splitting".

https://github.com/koalaman/shellcheck/wiki/SC2086

build-scripts/package
```diff
@@ -156,6 +156,29 @@ case "$PACKAGING" in
...
+        MOUNT_DIR=cfengine.img
+        mkdir $MOUNT_DIR
+        sudo mount $LOOP_DEV $MOUNT_DIR
+        sudo cp -pr var/cfengine/* $MOUNT_DIR
+        sudo umount $LOOP_DEV
+        sudo losetup --detach $LOOP_DEV
+	rmdir $MOUNT_DIR
+	# create "nested" tarball containing only *.img file
+	tar czf $P/coreos/cfengine3.img.tar.gz --sparse -C $(dirname $IMAGE) $(basename $IMAGE)
+	rm $IMAGE
```
vpodzime: Why is this nested tarball needed?

packaging/cfengine-nova/coreos/install.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+if [[ $EUID -ne 0 ]]; then
```
vpodzime: `[[ ]]` syntax is non-POSIX and the hashbang says `/bin/sh`.

packaging/cfengine-nova/coreos/install.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# extract image to expected location
+tar xf cfengine3.img.tar.gz -C /var
+
+# setup systemd
+cd systemd
+chmod 664 *
+cp * /etc/systemd/system/
+systemctl daemon-reload
+ls | xargs -l systemctl enable
```
vpodzime: `ls -1` would be safer here, I think

packaging/cfengine-nova/coreos/install.sh
```diff
@@ -0,0 +1,23 @@
...
+systemctl start var-cfengine.mount
+systemctl start cfengine3
```
vpodzime: Only `systemctl start cfengine3` should be needed here, the `.mount` service should be a requirement.

packaging/cfengine-nova/coreos/systemd/var-cfengine.mount
```diff
@@ -0,0 +1,14 @@
...
+[Install]
+WantedBy=cf-execd.service
+WantedBy=cf-monitord.service
+WantedBy=cf-serverd.service
+
+[Mount]
+What=/var/cfengine3.img
+Where=/var/cfengine
+Type=xfs
+Options=loop
```
vpodzime: We either need to add `discard` here or also ship a `.timer` service periodically running `fstrim /var/cfengine` to free-up space in the sparse image.

packaging/cfengine-nova/coreos/install.sh
vpodzime: And then it's not necessary to `enable` the service, I think.

packaging/cfengine-nova/coreos/install.sh
craigcomstock: `$EUID` is in bash not `/bin/sh` so probably need to use a different shebang or a different "root detection" method?

build-scripts/package
Lex-2008: Because I don't feel good about leaving 100Gb sparse file in random location (probably somewhere inside `/home`). Some utilities (`cp` and `rsync`) don't check for sparseness by default - so, for example, if you remove this `rm $IMAGE` line - our tests will fail, since they rsync everything form `/var/cfengine` to chroot environment, and it explodes 100Gb sparse file into a 100Gb simple file, and our test VMs have only 25Gb disk space… So if someone creates a backup of `/home` dir by rsyncing files - they might get an unexpected surprise.

packaging/cfengine-nova/coreos/install.sh
Lex-2008: `systemctl start var-cfengine.mount` blocks until the FS is mounted, while `systemctl start cfengine3` does not.

Regarding enabling service - I'll check, thanks!

build-scripts/package
Lex-2008: the line is `[ -d $PKG-* ] && cd $PKG-*`. Indeed, I didn't expect to have more than one `$PKG-*` dir. I'll consider adding a check that there is not more than one such dir.

build-scripts/package
Lex-2008: Agreed, although we don't do it in the whole file

packaging/cfengine-nova/coreos/install.sh
Lex-2008: hmm, in all my testing it always was defined:
```
-bash-4.2$ /bin/sh
sh-4.2$ echo $EUID
1011
sh-4.2$ exit
```
or was I testing it improperly?

Right, I guess we can use just 

```
if [ "`id -u`" != 0 ]; then
```

craigcomstock: Odd. What is /bin/sh for you though? Maybe it is bash? Mine is /bin/dash actually which is a classic debian-ism. ;)

Lex-2008: my bad, I was testing on CentOS and CoreOS :-)

Lex-2008:
Notes to self:
* uninstall
* replace `xargs` with `for … in … do`

packaging/cfengine-nova/coreos/install.sh
Lex-2008: we need to enable at least `cfengine3` service to ensure it's started on boot

Lex-2008:
rebuild: https://ci.cfengine.com/job/pr-pipeline/2511/

packaging/cfengine-nova/coreos/install.sh
vpodzime: > `systemctl start cfengine3` does not

I think it blocks if you add `After=var-cfengine.mount` together with `Requires`.

vpodzime: > we need to enable at least `cfengine3` service to ensure it's started on boot

Yeah, sure, I meant the `.mount` service.

build-scripts/package
vpodzime: Good point!

packaging/cfengine-nova/coreos/install.sh
Lex-2008: Right, I agree with adding the .mount service as an explicit dependency of the cfengine3 service.
However, note that as mentioned elsewhere, cfengine3 service can either wait for all subservices to start on `systemctl start cfengine3`, or wait for all subservices to stop on `systemctl stop cfengine3`. Which of these do we prefer?

vpodzime: `stop` needs to be synchronous. But here `After=var-cfengine.mount` will make sure that the mount is done before the CFEngine services started, AFAICT.

Lex-2008: Ok, so with `After=var-cfengine.mount`, start is synchronous. But we still need both `systemctl stop cfengine3` and `systemctl stop var-cfengine.mount` in `uninstall.sh` for `rmdir /var/cfengine` to succeed at the end.

Lex-2008:
rebuild: https://ci.cfengine.com/job/pr-pipeline/2534/

packaging/cfengine-nova/coreos/install.sh
vpodzime: > But we still need both `systemctl stop cfengine3` and `systemctl stop var-cfengine.mount` in `uninstall.sh` for `rmdir /var/cfengine` to succeed at the end.

Yes, and here the fact that `systemctl stop cfengine3` is synchronous (waits) comes handy. :)

packaging/cfengine-nova/coreos/install.sh
```diff
@@ -0,0 +1,22 @@
...
+cd systemd
+chmod 664 *
+cp * /etc/systemd/system/
+systemctl daemon-reload
+systemctl enable cfengine3
+systemctl start cfengine3
+
+# prepare to bootstrap
+/var/cfengine/bin/cf-key
+
```
vpodzime: The script should `cd` back to the original `$PWD`, I have scripts that move me to a different directory when I run them.

packaging/cfengine-nova/coreos/uninstall.sh
```diff
@@ -0,0 +1,22 @@
...
+cd systemd
+for service in *; do
+  rm /etc/systemd/system/$service
+  test -d /etc/systemd/system/$service.wants && rm -rf /etc/systemd/system/$service.wants
+done
+systemctl daemon-reload
+
+# delete extracted image
+rm /var/cfengine3.img
+rmdir /var/cfengine
```
vpodzime: Same here with `cd $ORIG_PWD` required.

packaging/cfengine-nova/coreos/install.sh
Lex-2008: weird, I always assumed it's the case only when _sourcing_ files, but not when _running_ them

vpodzime: Oh, could be the case! ~Scratch this, then.~

PR #587, opened 2019-07-31
ENT-4889 Adapt travis script to updated s3cmd

Lex-2008:
New version of `s3cmd` asks one more question to which we accept a default answer


cf-bottom:

PR #590, opened 2019-08-09
Override OPENSSLDIR in mingw builds

vpodzime:
We need it to be a Windows path. However, when building and
installing files, we need it to be a UNIX path (the build runs on
Debian). So let's just override the macro definition.

### DIFF
--- a/deps-packaging/openssl/mingw/debian/rules
+++ b/deps-packaging/openssl/mingw/debian/rules
@@ -12,9 +12,11 @@ clean:
 
 # AFAIK it is either i686-w64-mingw32 or x86_64-w64-mingw32
 ifeq ($(DEB_HOST_GNU_TYPE),i686-w64-mingw32)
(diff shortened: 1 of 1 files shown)


cf-bottom:

deps-packaging/openssl/mingw/debian/rules
```diff
@@ -13,8 +13,10 @@ clean:
 # AFAIK it is either i686-w64-mingw32 or x86_64-w64-mingw32
 ifeq ($(DEB_HOST_GNU_TYPE),i686-w64-mingw32)
     TARGET=mingw
+	CFENGINE_DIR=C:/Program Files/Cfengine
```
Lex-2008: is tab intentional here?

vpodzime: Actually, I think the two above lines shouldn't be indented at all. I'll change that.

PR #597, opened 2019-08-19
ENT-4800 custom workdir

Lex-2008:


cf-bottom:

Lex-2008:
@cf-bottom build, please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2604/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2604/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2604/

Lex-2008:
ready to be reviewed! Vendor-specific prefix is in the "ENT-4800 switch to different BUILDPREFIX" commit, which will be dropped before merging this.

deps-packaging/pkg-build-solaris
```diff
@@ -4,11 +4,20 @@ gen_prototype() {
...
+  PREPREFIX=${BUILDPREFIX%/*}
+  # PREPREFIX is parent dir for BUILDPREFIX - i.e. /var or /usr/localcw/opt
+  # all its dirs should be excluded - in second case, they are:
+  # * /usr
+  # * /usr/localcw
+  # * /usr/localcw/opt
+  # To check if $destf is substring of $PREPREFIX (or rather if $PREPREFIX
+  # starts with $destf), we extract first N characters from $PREPREFIX, where
+  # N is length of $destf, i.e. ${#destf}
   for f in `find $pkgdir`; do
```
vpodzime: Wouldn't it be easier to take the first dir in the path (e.g. `/usr`) and use `find -prune`?

build-scripts/clean-buildmachine
```diff
@@ -7,9 +7,9 @@ uninstall_cfbuild
 
 case $OS in
 aix)
-sudo rm -rf /var/cfengine/*
+sudo rm -rf $PREFIX/*
```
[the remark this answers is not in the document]
Lex-2008: I'm trying to add quotes all over our build infra, but it feels that allowing spaces in `$PREFIX` would be a bit bigger task. For example, what's the proper way of quoting this?
```
./configure --prefix=$PREFIX CPPFLAGS=-I$PREFIX/include
```
As I understand, `CPPFLAGS` can contain multiple space-separated arguments (like `CPPFLAGS=-I$PREFIX/include -Wall`), and somewhere later they become separate arguments to `gcc`.

I'm using this test script to emulate that:
```
gcc ()
{
	echo "gcc called with following arguments: [$1] [$2] [$3]"
}

export "$1" #quickly assign
echo "CPPFLAGS is [$CPPFLAGS]"

gcc $CPPFLAGS
```
and all my attempts to properly escape space fail:
```
bash test.sh CPPFLAGS="-I\"spaced path\" -Wall"
CPPFLAGS is [-I"spaced path" -Wall]
gcc called with following arguments: [-I"spaced] [path"] [-Wall]
```

Expected output is `gcc called with following arguments: [-I"spaced path"] [-Wall] []`, but I don't know how to achieve it.

Or am I missing something?

build-scripts/package
```diff
@@ -307,7 +305,7 @@ case "$PACKAGING" in
        cat tmp >> CONTENTS
        rm plist-head tmp
        mv CONTENTS ./+CONTENTS
-       tar cjvf "cfengine-nova-$VERSION\_1.tbz" +CONTENTS +DESC +COMMENT var/
+       tar cjvf "cfengine-nova-$VERSION\_1.tbz" +CONTENTS +DESC +COMMENT *
```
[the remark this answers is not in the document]
Lex-2008: It shouldn't. But anyway, this for freebsd, which we don't build for.

deps-packaging/php/debian/cfbuild-php-devel.install
```diff
@@ -1 +1 @@
-var/cfengine/httpd/php/include
+/var/cfengine/httpd/php/include
```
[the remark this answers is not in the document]
Lex-2008: You're right, install files are supposed to have relative paths: https://www.debian.org/doc/manuals/maint-guide/dother.en.html#install - but everywhere in our buildscripts they are absolute:
```
~/git/buildscripts/deps-packaging$ ls */debian/*.install | xargs head -qn1
/var/cfengine/httpd/build
/var/cfengine/httpd/bin
/var/cfengine/include
/var/cfengine/lib/apr-util-1/*.so
/var/cfengine/include
/var/cfengine/lib/*.so.*
/usr/bin/autoconf
/usr/bin/aclocal
/var/cfengine/bin/git*
/usr/bin/gendesc
/var/cfengine/include
/var/cfengine/bin/getfacl
/usr/include/attr
/var/cfengine/lib/*.so.*
/var/cfengine/bin/curl-config
/var/cfengine/bin/curl
/var/cfengine/include
/var/cfengine/lib/
/usr/bin/libtool
/var/cfengine/bin/xml2-config
/var/cfengine/lib/*.so.*
/var/cfengine/include/
/var/cfengine/bin
/var/cfengine/include/
/var/cfengine/bin/
/var/cfengine/include
/var/cfengine/lib/*.so.*
/var/cfengine/include
/var/cfengine/lib/libssl.so.*
/var/cfengine/bin/pcre-config
/var/cfengine/lib/*.so.*
/var/cfengine/httpd/php/include
/var/cfengine/httpd/php/bin
/var/cfengine/include/*
/var/cfengine/bin/clusterdb
/var/cfengine/bin/rsync
/var/cfengine/include
/var/cfengine/lib/libz.so.*
```

and later in {{install-dependencies}} we replace {{/var/cfengine}} with actual {{$BUILDPREFIX}}:
```
if [ "$BUILDPREFIX" != "/var/cfengine" -a \( "$DEP_PACKAGING" = "deb" -o "$ROLE" = hub \) ]
then
    # Apply necessary changes for alternate prefix
    sed -i "s:/var/cfengine:$BUILDPREFIX:"  \
        $BASEDIR/buildscripts/deps-packaging/*/debian/*.install  
```
(~current~ _old_ version you can see here: https://github.com/cfengine/buildscripts/blob/master/build-scripts/install-dependencies#L95)

deps-packaging/pkg-build-solaris
Lex-2008: I'm not sure I understand you. We need to process both files and dirs in current dir (recursively), excluding these three: `usr` `usr/localcw` `usr/localcw/opt` (but include their subdirs, like `usr/localcw/opt/cfengine`). What's here to prune? :-)

build-scripts/clean-buildmachine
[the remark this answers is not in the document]
vpodzime: `-Wall` is a `CFLAG` not a `CPPFLAG`. And yes, quoting of these things is really tricky. You can see how it has to be done at https://github.com/cfengine/buildscripts/pull/590/files#diff-a178f3bd31f3d53b413a93b1a97d9a5aR30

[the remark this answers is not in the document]
vpodzime: > I'm trying to add quotes all over our build infra, but it feels that allowing spaces in `$PREFIX` would be a bit bigger task.

I agree that it would be a big task. But no need to add more problematic places in this PR than we already have.

deps-packaging/php/debian/cfbuild-php-devel.install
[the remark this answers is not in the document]
vpodzime: :+1:

deps-packaging/pkg-build-solaris
vpodzime: >  # PREPREFIX is parent dir for BUILDPREFIX - i.e. /var or /usr/localcw/opt
>  # all its dirs should be excluded

This wording confused me, I think. In general I think this needs a bugger comment on why this is done and how it works -- because we don't provide info for those directories while we provide info about their subdirectories and files, right?

build-scripts/clean-buildmachine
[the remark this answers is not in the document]
Lex-2008: ok!

deps-packaging/pkg-build-solaris
Lex-2008: Right. These dirs (at least `/usr`) are provided by other packages, and adding them to our package with _conflicting_ perms breaks Solaris package manager.

Lex-2008:
@cf-bottom build on exotics

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/2741/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/2741/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-2741/

build-scripts/clean-buildmachine
```diff
@@ -7,9 +7,9 @@ uninstall_cfbuild
 
 case $OS in
 aix)
-sudo rm -rf /var/cfengine/*
+sudo rm -rf "$PREFIX"/*
 ;;
 *)
-sudo rm -rf /var/cfengine
+sudo rm -rf "$PREFIX"
```
craigcomstock: These just feel dangerous even on a buildmachine. Maybe be tricky and check to make sure that $(readlink -f "$PREFIX") != "/" ? Or some other such safety against rm -rf'ing / ?

build-scripts/functions
```diff
@@ -79,9 +79,13 @@ export_variables ()
 
 
     # BUILDPREFIX: which directory the packages will be installed to.
-    BUILDPREFIX=${BUILDPREFIX:-/var/cfengine}
+    BUILDPREFIX=${BUILDPREFIX:-/usr/localcw/opt/cfengine3}
```
craigcomstock: Why is the BUILDPREFIX changing always to not /var/cfengine?

build-scripts/install-dependencies
```diff
@@ -135,7 +135,11 @@ do
     esac
 
     revision=$(cat $BASEDIR/buildscripts/deps-packaging/revision)
-    version=0+$revision$versuffix$crossver
+    if [ "$BUILDPREFIX" != "/var/cfengine" ]
+    then
+        safe_prefix="+$(echo "$PREFIX" | sed 's:/::g')"
```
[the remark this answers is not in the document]
craigcomstock: https://www.debian.org/doc/debian-policy/ch-controlfields.html#version

> The upstream_version may contain only alphanumerics [6] and the characters . + - ~ (full stop, plus, hyphen, tilde) and should start with a digit. If there is no debian_revision then hyphens are not allowed.

build-scripts/travis
```diff
@@ -2,6 +2,8 @@
 
 set -ex
 
+PREFIX=/usr/localcw/opt/cfengine3
```
craigcomstock: Not sure why the default should be this non-standard PREFIX value. I understand testing a non-standard value also but seems like we should have the default here.

Lex-2008: This change (and a similar above) are in a separate commit, which will be dropped before merging. They are here just to test that it works at all :-)

build-scripts/clean-buildmachine
Lex-2008: sounds reasonable, although we might as well want to protect `/bin`, `/usr`, etc, but **NOT** `/cfe`, for example.

Lex-2008:
latest build: https://ci.cfengine.com/job/pr-pipeline/2852/

Lex-2008:
it seems that exotics don't understand fancy quotes - reverted, build retriggered:
https://ci.cfengine.com/job/pr-pipeline/2868/

Lex-2008:
Commit 56b795c "ENT-4800 Add quotes in deps-packaging" broke build on exotics, so I'm going to drop it. After reverting it, build is green: https://ci.cfengine.com/job/pr-pipeline/2868/ (with except of deployment tests which are expected to break since have `/var/cfengine` prefix hard-coded)

Lex-2008:
@vpodzime : please have a look again

PR #612, opened 2019-08-30
RHEL 8 support

vpodzime:
Small bits and pieces to make builds on RHEL 8 possible and resulting **agent** (for now) packages working.

Merge together:
https://github.com/cfengine/core/pull/3813
https://github.com/cfengine/masterfiles/pull/1492
https://github.com/cfengine/buildscripts/pull/612

### DIFF
--- a/build-scripts/configure
+++ b/build-scripts/configure
@@ -62,6 +62,10 @@ case "$WITH_SYSTEMD" in
   *)    var_append ARGS "--without-systemd-service" ;;
 esac
 
+# RHEL 8 requires an SELinux policy
+if [ "x$OS" = "xrhel" ] && [ "x${VER:0:1}" = "x8" ]; then
+  var_append ARGS "--with-selinux-policy"
+fi
 
 # Cross-compiling Windows?
 case "$ARCH-${OS_FAMILY}" in
--- a/deps-packaging/git/cfbuild-git.spec
+++ b/deps-packaging/git/cfbuild-git.spec
@@ -18,7 +18,7 @@ AutoReqProv: no
 mkdir -p %{_builddir}
 %setup -q -n git-%{git_version}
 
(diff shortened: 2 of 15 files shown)


cf-bottom:

deps-packaging/git/cfbuild-git.spec
```diff
@@ -18,7 +18,7 @@ AutoReqProv: no
 mkdir -p %{_builddir}
 %setup -q -n git-%{git_version}
 
-./configure --prefix=%{prefix} --with-openssl=%{prefix} --without-iconv --with-gitconfig=%{prefix}/config/gitconfig --with-gitattributes=%{prefix}/config/gitattributes --with-zlib=%{prefix} --with-curl=%{prefix}  --libexecdir=%{prefix}/lib
+./configure --prefix=%{prefix} --with-openssl=%{prefix} --without-iconv --with-gitconfig=%{prefix}/config/gitconfig --with-gitattributes=%{prefix}/config/gitattributes --with-zlib=%{prefix} --with-curl=%{prefix}  --libexecdir=%{prefix}/lib --with-python=/var/cfengine/bin/python
```
Lex-2008: For the sake of custom workdirs, can you use 
```
--with-python=%{prefix}bin/python
```
instead of hardcoded `/var/cfengine` value?

vpodzime: I surely should! Good catch! :+1:

build-scripts/configure
```diff
@@ -62,6 +62,10 @@ case "$WITH_SYSTEMD" in
   *)    var_append ARGS "--without-systemd-service" ;;
 esac
 
+# RHEL 8 requires an SELinux policy
+if [ "x$OS" = "xrhel" ] && [ "x${VER:0:1}" = "x8" ]; then
```
craigcomstock: Does rhel 7 also? I feel like it did when I was playing around with various VMs. Obviously we already have rhel 7 packages and such so you are almost certainly correct here. Just checking.

vpodzime: It's a bit weird and I didn't fully investigate it, to be honest. RHEL7 has SELinux in *Enforcing* by default and it also contains the incomplete upstream *cfengine* SELinux module. Yet everything works out of the box even without our extra module. I think the difference is that when *systemd* is doing the transition to the service it ends up in an unconfined domain. But I cannot tell for sure, as I said, I didn't spent much time on this.

PR #614, opened 2019-09-04
Patch LMDB to not abort on assert failure

vpodzime:
We want to have a chance to handle the failure gracefully.

Merge together:
https://github.com/cfengine/core/pull/3820
https://github.com/cfengine/buildscripts/pull/614

### DIFF
--- a/deps-packaging/lmdb/0001-Do-not-abort-if-a-user-define-assert-function-is-cal.patch
+++ b/deps-packaging/lmdb/0001-Do-not-abort-if-a-user-define-assert-function-is-cal.patch
@@ -1,0 +1,31 @@
+From 7b66c52e833f49a71a8a2900fb8452217395ef03 Mon Sep 17 00:00:00 2001
+From: Vratislav Podzimek <v.podzimek@mykolab.com>
+Date: Wed, 4 Sep 2019 16:25:53 +0200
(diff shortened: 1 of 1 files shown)


cf-bottom:

deps-packaging/lmdb/0001-Do-not-abort-if-a-user-define-assert-function-is-cal.patch
```diff
@@ -0,0 +1,31 @@
...
+index f685421..214a0eb 100644
+--- a/libraries/liblmdb/mdb.c
++++ b/libraries/liblmdb/mdb.c
+@@ -1766,8 +1766,10 @@ mdb_assert_fail(MDB_env *env, const char *expr_txt,
+ 		file, line, expr_txt, func);
+ 	if (env->me_assert_func)
+ 		env->me_assert_func(env, buf);
+-	fprintf(stderr, "%s\n", buf);
+-	abort();
++    else {
```
olehermanse: looks like indentation is wrong?

vpodzime: Yeah, but I don't care about their TABs.

PR #630, opened 2019-10-09
ENT-4754: Enabled nova repo to decide which tests to run

craigcomstock:
runtests.sh looks at TRAVIS environment variable and if present then skips tests which don't currently work there: migration and generate since they have dependencies which are not met.

merge with https://github.com/cfengine/nova/pull/1510


cf-bottom:

PR #638, opened 2019-11-14
ENT-5063 Fix state/pg perms

Lex-2008:


cf-bottom:

Lex-2008:
I believe we can't change permissions in the RPM because "cfpostgres" user might not exist while rpm extracts files before postinst script is running.

vpodzime:
> I believe we can't change permissions in the RPM because "cfpostgres" user might not exist while rpm extracts files before postinst script is running.

Ownership is one thing, permissions are another thing. The commands moved in this PR change both so I wonder if there is something wrong with the **permissions** in the RPM.

Lex-2008:
I see that you can have `%defattr(755,root,root,755)` in *.spec file, but I don't see anything similar in `debian/*.install` file. So let's keep it like this, and find a proper fix in a future - https://tracker.mender.io/browse/ENT-4267

PR #644, opened 2019-11-26
ENT-4981 Add build number to coreos and generic tarballs

vpodzime:
Only coreos tarballs are modified here in the 3.12.x port of the
original commit because generic tarballs don't even have the safe
prefix from commit 631125b7cba here.

(cherry picked from commit 8dc9c707574f2caba6c763b135d513f27996097f)

Conflicts:
  build-scripts/package


Lex-2008:
we don't have `$safe_prefix` here, because ENT-4800 wasn't ported to 3.12.x branch.

While technically it's safe (undefined variable `$safe_prefix` will just resolve to empty string), let's not add it here?

PR #665, opened 2020-01-15
Do not use '--with-tokyocabinet=/usr' when building tarballs

vpodzime:
We need to run './configure' to be able to run 'make dist' to get
a tarball. And './configure' requires one of LMDB or TokyoCabinet
to be installed (inlc. devel files). But we might as well just
install 'liblmdb-dev' on our bootstrap machines and let the
defaults do their job.


vpodzime:
jenkins build (no tests): [Build Status](https://ci.cfengine.com/job/pr-pipeline/4399/)

PR #669, opened 2020-01-27
Adapt inventory refresh in the postinstall script

vpodzime:
There is a new 'inventory_variables_refresh' task introduced in
cfengine/mission-portal@d2550582. We no longer need to do full
inventory refresh.
    
Ticket: ENT-4864
Changelog: None


cf-bottom:

PR #673, opened 2020-01-27
ENT-5115: Added checks for database state to avoid error messages during install (3.15.x)

craigcomstock:


cf-bottom:

PR #686, opened 2020-02-19
Fall back to 'pg_ctl stop -m fast' if '-m smart' fails

vpodzime:
The default timeout for stopping PostgreSQL with '-m smart'
(which tries to wait for connections to terminate) is 60
seconds. For some reason that is not always enough in the hub
postinstall scriptlet. So let's give it 120 seconds and fall back
to '-m fast' which terminates the connections (gracefully).


vpodzime:
Backports: https://github.com/cfengine/buildscripts/pull/687 and https://github.com/cfengine/buildscripts/pull/688

PR #717, opened 2020-03-27
Install Mission Portal Javascript dependencies fron NPM

aleksandrychev:
ChangeLog: None

Ticket: ENT-5454

### DIFF
--- a/build-scripts/travis
+++ b/build-scripts/travis
@@ -98,6 +98,18 @@ case "$JOB" in
 
         mv -T ldap-api mission-portal/ldap
 
+        # packages needed for installing Mission portal dependencies
+        sudo apt-get -y install npm
+        cd mission-portal/scripts
+        # install dependencies from npmjs
+        npm i
+        # make a build of jquery-ui from sources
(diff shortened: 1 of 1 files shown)


build-scripts/travis
```diff
@@ -98,6 +98,17 @@ case "$JOB" in
...
+        # pacakges needed for isntalling Mission portal dependecies
+        sudo apt-get -y install npm
+        cd mission-portal/scripts
+        # install dependincies from npmjs
+        npm i
+        # make a build of jquery-ui from sources
+        cd node_modules/jquery-ui
+        npm i
+        grunt sizer concat
+        rm -rf ./node_modules
```
Lex-2008: actually, you need to cd back **after** this command, because few lines below this file says:
```
./buildscripts/build-scripts/autogen
```
so it expects to be outside of mission-portal dir.

As I understand, you're 4 levels deep at this moment (`mission-portal/scripts/node_modules/jquery-ui`), so need to go 4 levels up

aleksandrychev: done

PR #726, opened 2020-04-11
Make sure version information is available when building tarballs

vpodzime:
So that the tarballs (some of which we also release) can have
proper names and version information.

Ticket: ENT-5429
Changelog: None

Merge together:
https://github.com/cfengine/masterfiles/pull/1737
https://github.com/cfengine/buildscripts/pull/726


cf-bottom:

PR #731, opened 2020-04-16
Use GCC 6 on new AIX 7.1 VMs

vpodzime:
The new IBM CECC VMs have GCC 8 as the default compiler
(yum install gcc), but it is broken or at least doesn't work
properly with our specific codebase because it produces binaries
that fail in weird ways. Thus we need to use GCC 6 (from the
'gcc6' package), but that is installed under a non-standard path
which we thus need to add to $PATH for things to work.

Ticket: ENT-5418


cf-bottom:

PR #737, opened 2020-04-30
Add EVP_KDF API to OpenSSL we build for RHEL

vpodzime:
This API is needed by PAM on RHEL 8.

Ticket: ENT-5564
Changelog: None


craigcomstock:
Would it be nice to add some info on where the patch came from and how it was generated? So that we have a process/instructions that we now need to track both upstream as well as redhat version of openssl?

vpodzime:
> Would it be nice to add some info on where the patch came from and how it was generated?

Very good point! Added to the commit message.

vpodzime:
Merging this to get greener builds ASAP. We can easily revert this if we choose a different approach.

PR #755, opened 2020-06-18
Travis: Attempt to fix broken 3.15.x builds

olehermanse:
Cherry-pick to bring 3.15.x up to date with master, since this part looked suspicious. On 3.15.x, `cf-secret` path was not specified in argument to testall, while on master, it didn't need to be, because it's using bindir instead.


olehermanse:
Let's see if this passes:
https://travis-ci.com/github/cfengine/mission-portal/builds/170246335

PR #756, opened 2020-06-19
ENT-6036 disable old services / master

Lex-2008:
Issue was that sometimes systemd doesn't remove service from dependency list
automatically. As result, we were left with a broken state when cfengine3
service depends on cf-consumer service which does not exist. Disabling the
service ensures that it's gone from dependency lists of all other services.

----

#

### DIFF
--- a/packaging/common/cfengine-hub/preinstall.sh
+++ b/packaging/common/cfengine-hub/preinstall.sh
@@ -127,6 +127,16 @@ if is_upgrade; then
   fi
 fi
 
+# ENT-6036: disable old services
+if is_upgrade; then
+    if [ -x /bin/systemctl ] && [ -e /usr/lib/systemd/system/cf-consumer.service ]; then
+        /bin/systemctl disable cf-consumer.service
+    fi
+    if [ -x /bin/systemctl ] && [ -e /usr/lib/systemd/system/cf-redis-server.service ]; then
+        /bin/systemctl disable cf-redis-server.service
+    fi
+fi
+
 if migrating_postgres; then
   cf_console echo "Moving old data and copying old binaries to $BACKUP_DIR"
   # Now that PostgreSQL is shut down, move the old data out of the way.


cf-bottom:

vpodzime:
> Issue was that sometimes systemd doesn't remove service from dependency list automatically.

`systemctl daemon-reload` should always do that. If not, it's an indication of some other problem, I think. Are you sure that there really is a failing dependency? Or is it just our scriptlet trying to start all the services running before the upgrade? If the latter, we should just modify that part of the scriptlet.

> Disabling the service ensures that it's gone from dependency lists of all other services.

That's not what `systemctl disable` does, AFAICT.

Lex-2008:
Without this fix (and relevant quickfix in triple-upgrade test, https://github.com/cfengine/system-testing/pull/342/commits/7e14fb3628e80317909c7fe230883b0fe3146292), triple-upgrade test fails like this:

https://ci.cfengine.com/view/all/job/sequential-tests/550/testReport/junit/(root)/new_tests_sequential_tests_05-upgrade-triple/01_upgrade_triple_test/

vpodzime:
> Without this fix (and relevant quickfix in triple-upgrade test, [cfengine/system-testing@7e14fb3](https://github.com/cfengine/system-testing/commit/7e14fb3628e80317909c7fe230883b0fe3146292)), triple-upgrade test fails like this:

```
+ echo "Errors when installing 3.12.5 version HUB packages:"
Errors when installing 3.12.5 version HUB packages:
+ grep -P "(err|fail)" .13465.install.log.filtered
34.246.177.23	Failed to start cf-consumer.service: Unit cf-consumer.service failed to load: No such file or directory.
34.246.177.23	Failed to start cf-redis-server.service: Unit cf-redis-server.service failed to load: No such file or directory.
```
to be more precise.

So yes, I think this just comes from the `get_cfengine_state` and `restore_cfengine_state` in the scriptlets. However, there is [a mechanism](https://github.com/cfengine/buildscripts/blob/master/packaging/common/script-templates/script-common.sh#L57) that should prevent that from happening. I guess that's what needs an improvement/fixup.

Lex-2008:
> However, there is [a mechanism](https://github.com/cfengine/buildscripts/blob/master/packaging/common/script-templates/script-common.sh#L57) that should prevent that from happening. I guess that's what needs an improvement/fixup.

Right, and the issue is that it's [not there on 3.12.x branch][t]!
[t]: https://github.com/cfengine/buildscripts/blob/3.12.x/packaging/common/script-templates/script-common.sh#L56

Lex-2008:
proper fix in https://github.com/cfengine/buildscripts/pull/759

PR #763, opened 2020-07-27
ENT-5363 add SUSE 15 to Jenkins

Lex-2008:


cf-bottom:

Lex-2008:
green build: https://ci.cfengine.com/job/pr-pipeline/5681/

build-scripts/test-on-testmachine
```diff
@@ -46,6 +46,12 @@ case "$TEST_MACHINE" in
             then
                 continue
             fi
+            # gpg-agent sometimes is spawned on suse during our tests.
+            # This happens only in chroot. Ignore that.
+            if ps -ef | egrep "^ *[^ ]+ +$pid" | grep gpg-agent && grep suse /etc/os-release
```
vpodzime, line 51: A couple things here:

- first test `grep suse /etc/os-release`
- `ps -e -o FIELDS` might be better to get the info we need/want and make parsing easier
- we should still kill those processes, right? `ps -o cmd` should be possible to use to find the *gpg-agent* processes spawned for zypper. `ps --forest` might even be possible to use to find the processes spawned from the tests

vpodzime (commented):
Looks good to me otherwise.

build-scripts/test-on-testmachine
Lex-2008: addressed first two points, regarding the third one - yes, later in the code all processes running in chroot are killed anyway, with a message like this:

```
                echo "Error: Found processes left behind in the chroot. I'm killing them, but the build will fail, so clean up your mess!"
```

vpodzime (approved):
Looks good to me.

olehermanse (approved):

PR #778, opened 2020-10-07
Change DEBUG builds to no optimizations (-O0) to make gdb usage easier

craigcomstock:


craigcomstock:
Wow. For just one windows package this makes the msi file around 700MB :O so, that's fun and probably not something we want to generate a lot of? Maybe a different `-O` option is smaller but still useful?

vpodzime:
>  Maybe a different `-O` option is smaller but still useful?

Yeah, we should try and see what `-O0` yields.

craigcomstock:
I suspect that the problem on windows with getting a back trace had something other than -g or optimization settings. I will drop this one for now until another time.

PR #795, opened 2021-01-14
ENT-6529 see if bootstrap-tarballs succeeds with -e, in case masterfiles tar-package fails sometimes

craigcomstock:
try together
https://github.com/cfengine/buildscripts/pull/795
https://github.com/cfengine/masterfiles/pull/1893

### DIFF
--- a/build-scripts/bootstrap-tarballs
+++ b/build-scripts/bootstrap-tarballs
@@ -1,4 +1,4 @@
-#!/bin/sh -x
+#!/bin/sh -x -e
 
 . `dirname "$0"`/functions
 . detect-environment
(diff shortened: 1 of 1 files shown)


craigcomstock:
@cf-bottom jenkins please

craigcomstock:
@Lex-2008 what do you think about this? I LOVE that `-e` doesn't fail a build! :)

https://ci.cfengine.com/job/bootstrap-pr/7315/console

I was thinking it would be nice to check assumptions after an operation.

Ideally I'd like to make assertions before hand too... so downstream from here we could check for the mpf tarball at just the right moment to ensure that say net transfer succeeded or similar.

If you like this, I can re-work into a proper PR. :)

craigcomstock:
other options resolved this issue.

PR #797, opened 2021-02-03
Switched zlib to mirror

Lex-2008:
Issue was that zlib.net wasn't working for Oslo-DC

----

#


olehermanse (changes requested):
Missing newline at EOF. Also, according to vratislav, it should work tomorrow since ubuntu 16 is offline, and others have cache.

vpodzime (commented):
The URL works :+1:

PR #798, opened 2021-02-05
DEBUG: Add asserts to LMDB to find out where EINVAL comes from

vpodzime:


cf-bottom:

vpodzime:
Unfortunately, even these changes are not helpful enough for identification of the root cause of the issues.

PR #805, opened 2021-02-18
DEBUG: Do not override options from --enable-debug

vpodzime:
Because they actually produce binaries that can be debugged.

### DIFF
--- a/build-scripts/configure
+++ b/build-scripts/configure
@@ -82,9 +82,6 @@ case "$BUILD_TYPE" in
     ;;
   DEBUG)
     ARGS="$ARGS --enable-debug"
-    # Override the default "-g3 -O0" that comes with ./configure --enable-debug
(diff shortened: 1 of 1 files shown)


cf-bottom:

Lex-2008 (approved):

vpodzime:
I wonder if we should actually merge this. The resulting packages and binaries are much bigger, but do we need to care?

Lex-2008:
yeah, we do care for nightly and pr-pipeline builds, 'cause we're paying for traffic. Do I understand it right that it will affect our nightlies and pr-pipelines? Can we make it optional - like "super-debug" builds?

vpodzime:
> Can we make it optional - like "super-debug" builds?

Yeah, we could do something like that. But that's a lot more work than just creating a simple PR like this and including it in a build.

PR #809, opened 2021-03-24
Fix expected zlib tarball checksum

vpodzime:
sha256sum result on a tarball downloaded over HTTPS so this must
be the correct one.


vpodzime:
Scratch this, please, I was getting an HTML document not a tarball.

PR #814, opened 2021-04-01
CFE-3518: Removed unused plugins directory (master)

nickanderson:
merge together: https://github.com/cfengine/masterfiles/pull/1956

The plugins directory has gone unused since it's introduction in 2013. This
change cleans up the plugins directory if it exists and is not empty. If it is
not empty the user is informed and asked to review their policy.

Ticket: CFE-3518
Changelog: Title


vpodzime (approved):
ACK

PR #840, opened 2021-06-11
ENT-6101 drop unsupported platforms

Lex-2008:
And add Ubuntu 20


cf-bottom:

craigcomstock (approved):

build-scripts/package
```diff
@@ -249,7 +249,7 @@ case "$PACKAGING" in
     (cd "$BASEDIR/$PKG/pkg"; export DEB_BUILD_OPTIONS="$DEB_BUILD_OPTIONS"; dpkg-buildpackage -b -us -uc -rfakeroot) || false
 
     # Create TAR package
-    TARBALL="$BASEDIR/$PKG/pkg/$PKG-$DEB_VERSION$safe_prefix.$OS$os_version_major.$ARCH.pkg.tar.gz"
+    TARBALL="$BASEDIR/$PKG/pkg/$PKG-$DEB_VERSION$safe_prefix.$os_version_major.$ARCH.pkg.tar.gz"
```
vpodzime, line 252: Shouldn't `$os_version_major` be dropped as well?

vpodzime (changes requested):
The first commit message of the first commit is wrong. It should use `<` instead of `<=` and also, the commit leaves Ubuntu 16 in the list(s) so clearly `Ubuntu < 16` is unsupported.

vpodzime (approved):
Looks good to me now.

PR #845, opened 2021-06-30
Compiled Mission portal styles in bootstrap jobs

aleksandrychev:
Ticket: ENT-6007

### DIFF
--- a/build-scripts/bootstrap-tarballs
+++ b/build-scripts/bootstrap-tarballs
@@ -109,6 +109,13 @@ fi
 )
 
 (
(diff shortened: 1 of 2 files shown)


craigcomstock (commented):
this is good but I'd like you to test a change which breaks the less->css compilation and make sure that errors come out in logs in CI if possible so it is easy to see what went wrong and fix.

craigcomstock (approved):
awesome. saw the failure messages in slack that you tested with broken less :+1: 🚀

PR #852, opened 2021-07-30
Add localhost for 127.0.0.1 in /etc/hosts in test chroot

vpodzime:
So that connections from 127.0.0.1 appear as connections coming
from localhost with reverse lookup.

Ticket: ENT-7362
Changelog: None


vpodzime:
This wasn't needed.

PR #864, opened 2021-10-20
ENT-7554: Changed 4th octet in Windows package version to match package release number (3.15)

nickanderson:


nickanderson:
@cf-bottom jenkins, please

PR #868, opened 2021-11-03
ENT-7898: Make sure files and dirs created in scriptlets have correct SELinux labels

vpodzime:
In particular, simple `mkdir` doesn't create the folder(s) with
correct SELinux label, unless the `-Z` option is used.

Ticket: ENT-7898
Changelog: VCS configuration now works on RHEL 8 hubs

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -1023,4 +1023,10 @@ fi
 
 rm -f "$PREFIX/UPGRADED_FROM.txt"
 
+# Let's make sure all files and directories created above have correct SELinux
+# labels.
+if command -v restorecon >/dev/null; then
+  restorecon -iR /var/cfengine /opt/cfengine
+fi
+
 exit 0
--- a/packaging/common/cfengine-non-hub/postinstall.sh
+++ b/packaging/common/cfengine-non-hub/postinstall.sh
@@ -140,4 +140,10 @@ fi
 
 rm -f "$PREFIX/UPGRADED_FROM.txt"
 
+# Let's make sure all files and directories created above have correct SELinux
+# labels.
+if command -v restorecon >/dev/null; then
+  restorecon -iR /var/cfengine /opt/cfengine
+fi
+
 exit 0


vpodzime:
jenkins: [Build Status](https://ci.cfengine.com/job/pr-pipeline/7253/)

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -1023,4 +1023,10 @@ fi
 
 rm -f "$PREFIX/UPGRADED_FROM.txt"
 
+# Let's make sure all files and directories created above have correct SELinux
+# labels.
+if command -v restorecon >/dev/null; then
+  restorecon -R /var/cfengine /opt/cfengine
+fi
```
nickanderson, line 1030: Makes sense.

packaging/common/cfengine-non-hub/postinstall.sh
```diff
@@ -140,4 +140,10 @@ fi
 
 rm -f "$PREFIX/UPGRADED_FROM.txt"
 
+# Let's make sure all files and directories created above have correct SELinux
+# labels.
+if command -v restorecon >/dev/null; then
+  restorecon -R /var/cfengine /opt/cfengine
+fi
+
```
nickanderson, line 148: makes sense.

packaging/common/cfengine-hub/postinstall.sh
nickanderson: TIL about `command` bash builtin. I don't see _bash_ shebang or anything, are we certain this will be executed by bash?

nickanderson (approved):
Seems ok to me, didn't know about `command -v`

packaging/common/cfengine-hub/postinstall.sh
vpodzime: Well, I think that's what we determined to be the most reliable thing to do because `which restorecon` didn't exit with non-zero code if the command was not found on some platforms. We use `command -v` in other places in these scriptlets, so should be fine here too.

vpodzime:
> jenkins: [Build Status](https://ci.cfengine.com/job/pr-pipeline/7253/)

rebuild with `-i` added to ignore non-existing files in `restorecon`: [Build Status](https://ci.cfengine.com/job/pr-pipeline/7257/)

PR #936, opened 2022-02-24
Improved enterprise hub package upgrade logging

nickanderson:
It's kind of nice to know why what's happening. Reading the log prior to this
change, you see that something is copied and deleted and copied and maybe
deleted again for why? This simply prints out some messages in the event of
failure so that it's a bit easier to read the logs.


craigcomstock (approved):
yes please!

PR #937, opened 2022-03-01
Check for existence of pg/data/postgresql.conf file, instead of data dir

Lex-2008:
Issue was that when /var/cfengine/state/pg/data is a mounted volume, it can't be deleted. First, it makes `rm -rf` fail, and second, it confuses migration logic which checks lack of this directory as a signal to create new postgresql database there (including postgresql.conf file). But when this directory is a mounted volume - `rm -rf` can delete everything inside it, except the directory itself. Empty directory is not a valid postgresql database, what confuses rest of postinstall script.

Note that we don't want to use plain is_upgrade here, because the generate_new_postgres_conf must be executed in two cases:
* on a fresh install
* when a major upgrade of PostgreSQL requires DB migration - happens only during minor CFEngine upgrade (3.15 -> 3.18)

and is_upgrade is, well, _any_ CFEngine upgrade.

Ticket: ENT-8423

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -368,7 +368,11 @@ init_postgres_dir()
   new_pgconfig_file="$1"
   pgconfig_type="$2"
 
-  rm -rf $PREFIX/state/pg/data
+  if test -e $PREFIX/state/pg/data; then
+    if ! rm -rf $PREFIX/state/pg/data; then
+      cf_console echo "Warning: $PREFIX/state/pg/data couldn't be deleted"
+    fi
+  fi
   mkdir -p $PREFIX/state/pg/data
   chown -R cfpostgres $PREFIX/state/pg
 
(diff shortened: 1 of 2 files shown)


Lex-2008:
@cf-bottom build plz

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/7637/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/7637/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-7637/

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -368,7 +368,11 @@ init_postgres_dir()
   new_pgconfig_file="$1"
   pgconfig_type="$2"
 
-  rm -rf $PREFIX/state/pg/data
+  if test -e $PREFIX/state/pg/data; then
```
craigcomstock, line 371: so a `test -e` will get around things being a mount and avoid the "busy" problem with `rm -rf`? 👍

craigcomstock (approved):

craigcomstock:
@Lex-2008 definitely clean up the commit comment though, and squash. :+1:

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: `test -e` is here so we don't print the warning message if there's nothing to delete.

What really prevents script from aborting when `pg/data` is a mount point is putting the `rm -rf` into `if`, which swallows its false-y return code.

I could also just write `rm -rf ... || true`, but thought that printing a warning would be nicer.

vpodzime (approved):
Looks good to me.

PR #942, opened 2022-03-03
ENT-7505 Apply cfdb schema to hub_0 schema in case of superhub

craigcomstock:
Ticket: ENT-7505
Changelog: title


aleksandrychev:
@craigcomstock no need to apply schema changes to hub_0 in case of superhub, when you apply to public (main tables), then all partitions will get this change.

PR #944, opened 2022-03-10
Name install logs uniquely

Lex-2008:
Ticket: ENT-8342

### DIFF
--- a/packaging/common/script-templates/script-common-header-last.sh
+++ b/packaging/common/script-templates/script-common-header-last.sh
@@ -30,9 +30,12 @@ case "`os_type`" in
         ;;
 esac
 
-INSTLOG=/var/log/CFEngine-Install.log
+INSTLOG="/var/log/CFEngine-Install-$(date '+%Y-%m-%d_%H:%M:%S_%Z').log"
 mkdir -p "$(dirname "$INSTLOG")"
 touch "$INSTLOG"
+rm -f /var/log/CFEngineHub-Install.log
+rm -f /var/log/CFEngine-Install.log
+ln -s "$INSTLOG" /var/log/CFEngine-Install.log
 chown root:$INSTLOGGROUP "$INSTLOG"
 chmod 600 "$INSTLOG"
 CONSOLE=7


Lex-2008:
@cf-bottom build exotics plz

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/7697/)

(with exotics)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/7697/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-7697/

packaging/common/script-templates/script-common-header-last.sh
```diff
@@ -30,11 +30,17 @@ case "`os_type`" in
         ;;
 esac
 
-INSTLOG=/var/log/CFEngine-Install.log
+INSTLOG="/var/log/CFEngine-Install-$(date '+%Y-%m-%d_%H:%M:%S_%Z').log"
 mkdir -p "$(dirname "$INSTLOG")"
 touch "$INSTLOG"
+rm -f /var/log/CFEngineHub-Install.log
```
craigcomstock, line 36: why remove these? There should only be one, or two if a much older version was installed (CFEngineHub-Install).

And what about piling up date-named logs? I don't really care honestly, maybe just keep everything. There shouldn't be that many typically.

Lex-2008: re: why remove these - my line of thinking was in order not to confuse customers, who see CFEngine-Install.log and CFEngineHub-Install.log files next to each other, and think: "oh, I'm installing a Hub", so I probably need to look into the CFEngineHub-Install.log file". Or do you think it's better to keep the old CFEngineHub-Install.log file for the reference?

Note that currently, we use CFEngine-Install.log _file_, which will be _deleted_ during upgrade to the version with this change.

regarding piling up date-named logs - I've created [ENT-8452](https://tracker.mender.io/browse/ENT-8452) to fix it on exotics, and I think it's ok to keep them on the client's machines.

vpodzime: What about renaming that file to something that suggests it's an old log file? Like *CFEngine-Install-old.log*.?

vpodzime (commented):
Looks good to me, please drop the *test* commit.

packaging/common/script-templates/script-common-header-last.sh
Lex-2008: you mean, check if `/var/log/CFEngine-Install.log` is a file, and not symlink, and rename it to `/var/log/CFEngine-Install.log.old`? And `/var/log/CFEngineHub-Install.log` then should be moved to `/var/log/CFEngine-Install.log.very-old`.

But what if `/var/log/CFEngineHub-Install.log` exists, and  `/var/log/CFEngine-Install.log` does not (or is a symlink)? Should we then move `/var/log/CFEngineHub-Install.log` to `/var/log/CFEngine-Install.log.old`? But what if it already does exist?

Anyway, I think it's too much logic for too few gain.

vpodzime: > Anyway, I think it's too much logic for too few gain.

Agreed. Although, I only meant renaming `/var/log/CFEngineHub-Install.log` so that people are not confused by the word `Hub` in its name.

vpodzime:
> please drop the test commit

Still want this, please. :)

packaging/common/script-templates/script-common-header-last.sh
Lex-2008: That's right. But my concern is about someone who upgraded from 3.10 to 3.15: 3.10 logged to `CFEngineHub-Install.log` file, and 3.15 - to `CFEngine-Install.log`. If they upgrade to a version with this PR, then in its current state it will plainly delete the `CFEngine-Install.log` file (with details about 3.15 installation) - so I think that it might be more harm than good to keep a log of 3.10 installation.

Even worse if they upgraded from 3.10.x to 3.15.1 to 3.15.2 to 3.15.3 - then `CFEngineHub-Install.log` file has details about installation which happened _several_ upgrades before current one.

That what makes me think that it's better to remove that file so we don't have a chance to look at it :)

PR #958, opened 2022-03-30
ENT-8495: Stopped loading webdav apache modules by default

nickanderson:
Merge together:
- https://github.com/cfengine/buildscripts/pull/958
- https://github.com/cfengine/masterfiles/pull/2275

CFEngine Enterprise does not currently use webdav by default, so we do not need
to load these modules. They are still present in the build so users can still
leverage them if desired.

Ticket: ENT-8495
Changelog: CFEngine Enterprise Hub Apache no longer loads webdav modules by default


vpodzime (approved):
ACK

PR #961, opened 2022-04-11
ENT-8504: Stopped loading mod_info

nickanderson:
Merge together: 
- https://github.com/cfengine/masterfiles/pull/2285

Since we do not use the features provided by this module we don't need to load it
by default.

Ticket: ENT-8504
Changelog: Title


cf-bottom:

PR #962, opened 2022-04-11
ENT-8506: Stopped loading mod_status by default

nickanderson:
Merge together:
- https://github.com/cfengine/masterfiles/pull/2284

Since we don't make active use of this module, we don't need to load it by default.

Ticket: ENT-8506
Changelog: Title


cf-bottom:

PR #967, opened 2022-04-15
ENT-8498: Stopped loading Apache substitute module by default on Enterprise Hubs

nickanderson:
Since we do not use the functionality provided by this module, we should not
load it by default.

Ticket: ENT-8498
Changelog: Title


cf-bottom:

PR #970, opened 2022-04-15
ENT-8499: Stopped loading Apache mod_spelling by default on Enterprise Hubs

nickanderson:
We do not use the features provided by this module so we should not load it by
default.

Ticket: ENT-8499
Changelog: Title


cf-bottom:

PR #971, opened 2022-04-15
ENT-8499: Stopped loading Apache mod_spelling by default on Enterprise Hubs (3.18)

nickanderson:
We do not use the features provided by this module so we should not load it by
default.

Ticket: ENT-8499
Changelog: Title
(cherry picked from commit d7ae6384014be5d10a574cffe4fcb2bcedb2bb76)


cf-bottom:

PR #973, opened 2022-04-18
ENT-8497: Stopped loading Apache mod_version by default on Enterprise Hubs

nickanderson:
We do not use the functionality provided by this module, so we should  not load
it by default.

Ticket: ENT-8497
Changelog: Title


cf-bottom:

PR #974, opened 2022-04-18
ENT-8497: Stopped loading Apache mod_version by default on Enterprise Hubs (3.15)

nickanderson:
We do not use the functionality provided by this module, so we should  not load
it by default.

Ticket: ENT-8497
Changelog: Title
(cherry picked from commit 0ca6df74c6c5a967556fe002dd23af12bb6a69a6)


cf-bottom:

PR #977, opened 2022-04-18
ENT-8514: Stopped loading Apache mod_autoindex by default on Enterprise Hubs (3.18)

nickanderson:
default.

Ticket: ENT-8514
Changelog: Title
(cherry picked from commit 2a35b284118c85f8e1a5af8bef07b679dd0e33c2)


cf-bottom:

PR #1009, opened 2022-04-20
ENT-8517: Stopped loading Apache mod_mime_magic by default on Enterprise Hubs (3.18)

nickanderson:
default.

Ticket: ENT-8517
Changelog: Title
(cherry picked from commit 9f0f380c1ac3a7c2651d942e0a1cb344a74e400a)


cf-bottom:

PR #1028, opened 2022-04-25
ENT-8524: Stopped loading Apache mod_filter and mod_ext_filter by default on Enterprise Hubs

nickanderson:
We do not use the features provided by these modules, so we should not load them by default.


cf-bottom:

PR #1069, opened 2022-05-27
patch apachectl to be more resillient

Lex-2008:
Ticket?


Lex-2008:
cf-bottom: willit build?

PR #1082, opened 2022-06-17
Move ubuntu 22 arm builds out of exotics

craigcomstock:
They are fairly fast, cheaper than most instances in AWS,
and would be cool to provide them as nightlies to folks.

Ticket: none
Changelog: none


Lex-2008 (approved):

vpodzime (approved):
ACK

PR #1091, opened 2022-07-01
[php] ask openssl to ignore no EOL

Lex-2008:


cf-bottom:

deps-packaging/php/cfbuild-php.spec
```diff
@@ -6,6 +6,7 @@ Version: %{version}
 Release: 1
 Source0: php-%{php_version}.tar.gz
 Source1: php.ini
+Patch0: openssl-eol.patch
```
craigcomstock, line 9: why is it called "eol" patch and not "eof" or "openssl-ignore-unexpected-eof.patch" even to be horribly verbose?

Lex-2008:
indeed! Actually, that's a wrong PR and should be closed. Thanks for noticing and sorry for the noise!

@cf-bottom bad bot

cf-bottom:
I'm not sure I understand, @Lex-2008.

craigcomstock:
@cf-bottom I think what @Lex-2008 is trying to say to you is that this PR is invalid and should be closed. Right @Lex-2008 ? ;)

cf-bottom:
I'm not sure I understand, @craigcomstock.

PR #1099, opened 2022-07-14
TEST: all renames in 1 place

Lex-2008:


Lex-2008:
see DT failures in https://ci.cfengine.com/job/pr-pipeline/8294/

PR #1107, opened 2022-08-05
Take 64-bit library on sparc and 32-bit library elsewhere

Lex-2008:
Note that this assumes that we build only 64-bit binaries on sparc, and 32-bit binaries on other platforms. As of writing, this is correct ("other platforms" being only 32-bit x86 Intel).

Ticket: ENT-9013

Issue was that recently we switched to building 64-bit binaries on sparc (apparently they can run both 32 and 64-bit), but we still were packaging 32-bit library, and thus all our binaries were crashing on start.

### DIFF
--- a/deps-packaging/libgcc/solaris/build
+++ b/deps-packaging/libgcc/solaris/build
@@ -10,6 +10,12 @@ PREFIX=${BUILDPREFIX}
 #
 # Gross hack, but what we could do?
 #
-LIBGCC_LOC="/opt/csw/lib/"
+if [ "$(uname -p)" = "sparc" ]; then
+	# on 64-bit sparc CPUs, we take 64-bit library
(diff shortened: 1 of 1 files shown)


cf-bottom:

Lex-2008:
Manually tested running cf-agent produced by this build: https://ci.cfengine.com/job/pr-pipeline/8446/

Lex-2008:
Yes, it's only Solaris (OS), which can run on sparc (SunOS own CPU architecture), or on Intel x86 CPU architecture. I don't know if Solaris was/is/will be ported to other CPU architectures, and if we will support it or not - but for now we support Solaris on sparc and 32-bit x86 CPUs only.

vpodzime (approved):
ACK

PR #1110, opened 2022-08-11
Make sure liblmdb libraries are installed to /var/cfengine/lib not lib64

Lex-2008:
For some reason, on SUSE 15, liblmdb.so ends up in /var/cfengine/lib64.


cf-bottom:

Lex-2008:
it was decided to switch back to SLES instead of opensuse

PR #1113, opened 2022-08-31
CFE-4010: Revert "Workaround automake version problems between bootstrap host and ubuntu 22 build host"

craigcomstock:


craigcomstock:
run on ubu22 arm64 hub only [Build Status](https://ci.cfengine.com/job/pr-pipeline/8531/)

craigcomstock:
Another try. Revert the retry of make check and see if arm64hub ubu22 fails with automake version due for some reason to it wanting to reconfigure as if some Makefile.am or other item changed since config.status/etc were written. If it fails, then maybe check between config.status automake version (parsed) and automake --version and if mismatch, run `autoreconf -i`.

I didn't see any other way to check before-hand if there was a version mismatch.

Lex-2008 (approved):

PR #1119, opened 2022-09-06
ENT-9130: Fix masterfiles/tests/unit Makefile recreated during builds

craigcomstock:
Ticket: ENT-9130
Changelog: none


craigcomstock:
This doesn't occur on many platforms. Trying only with rhel 8: [Build Status](https://ci.cfengine.com/job/pr-pipeline/8551/)

craigcomstock:
marking this as invalid as the change/fix is in masterfiles automake files :)

PR #1126, opened 2022-09-26
Add an option to use system openssl library, instead of bundling - 3.18.x

Lex-2008:
Note that currently CFEngine requires OpenSSL version 1.1 minimum, so it won't work on older platforms like Debian 9.

Auto-detected based on $label env variable, which is set by Jenkins. To override autodetection, export $SYSTEM_SSL variable, values: 0 - do not use system openssl library, but build and bundle as usual 1 - use system openssl library

Ticket: ENT-8962

Changelog: none
(cherry picked from commit 31d0cb81507e3b327ad8c82233547c6a58501a57)


Lex-2008:
testbuild: https://ci.cfengine.com/job/pr-pipeline/8653/

PR #1130, opened 2022-10-11
Set cfbs 2.4.3 version that will be shipped together with hub packages

aleksandrychev:
Ticket: ENT-9456

### DIFF
--- a/packaging/cfengine-nova-hub/cfengine-nova-hub.spec.in
+++ b/packaging/cfengine-nova-hub/cfengine-nova-hub.spec.in
@@ -135,7 +135,7 @@ touch $RPM_BUILD_ROOT/var/log/postgresql.log
 %if %{?rhel}%{!?rhel:0} >= 7
 # Make sure cfbs is available and in place
 mkdir -p $RPM_BUILD_ROOT/tmp/cfbs_root
-pip3 install --root $RPM_BUILD_ROOT/tmp/cfbs_root cfbs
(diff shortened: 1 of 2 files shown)


olehermanse (approved):

olehermanse:
@Lex-2008 please comment on whether this is the "correct" way to do it, or if there's something like a variable or a specific file which should be used for version number.

olehermanse:
@cf-bottom jenkins, please.

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/8724/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/8724/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-8724/

Lex-2008 (approved):
well yeah, we could add it as a dependancy to deps-packaging sibdir, and create deb/rpm for it, in order not to depend on pip repos. But if we want to go that way - it would be a bigger task, deserving its own ticket.

Also, this way increases our download stats! :laughing: :chart_with_upwards_trend: 

Also, we need to remember to update this with every release. @aleksandrychev : will you do it as part of MP dependency update or is it better to create a separate ticket?

aleksandrychev:
@Lex-2008 I have created a release ticket https://tracker.mender.io/browse/ENT-9462, but yes would be better to have deps-packaging subdir with rules and not increase cfbs version manually on every release.

PR #1131, opened 2022-10-12
CFE-3807: Added assert on EINVAL in mdb_mutex_failed()

craigcomstock:
To debug issues related to the following log message:

error: Could not open database txn /var/cfengine/state/cf_lock.lmdb: Invalid argument

Ticket: CFE-3807
Changelog: title

related: 
https://github.com/cfengine/core/pull/5073
https://github.com/cfengine/buildscripts/pull/1131
https://github.com/cfengine/system-testing/pull/457

### DIFF
--- a/deps-packaging/lmdb/0011-assert-on-mdb-mutex-failed.patch
+++ b/deps-packaging/lmdb/0011-assert-on-mdb-mutex-failed.patch
@@ -1,0 +1,30 @@
+From d73eeaacb333792f6eb6c0b113d22daaa64dc05b Mon Sep 17 00:00:00 2001
+From: Craig Comstock <craig.comstock@northern.tech>
+Date: Tue, 27 Sep 2022 11:21:36 -0500
+Subject: [PATCH] Added assert in mdb_mutex_failed for rc != EINVAL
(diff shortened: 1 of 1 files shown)


Lex-2008 (approved):
Note to self: we apply all patches which filename starts with `00`.

@craigcomstock : so the error message will be more verbose and useful now?

craigcomstock:
> Note to self: we apply all patches which filename starts with `00`.
> 
> @craigcomstock : so the error message will be more verbose and useful now?

Yes. If a system encounters this "Could not open database txn  ... invalid argument" issue they can enable core dumps as appropriate for their system, install a debug-enabled build, and get a backtrace that might help us find the root cause of the issue.

Practically speaking this change is really only for us in our CI system. We do see this issue from time to time and since we run CI with debug enabled builds and with the https://github.com/cfengine/system-testing/pull/457 change always enable core dumps on deployment test machines, hopefully we will get a backtrace sooner than later ourselves.

craigcomstock:
This change has been through many iterations of CI and never failed anything BECAUSE of this change. So merged.

PR #1134, opened 2022-10-19
CFE-3807: Added assert on EINVAL in mdb_mutex_failed() (3.18)

craigcomstock:
To debug issues related to the following log message:

error: Could not open database txn /var/cfengine/state/cf_lock.lmdb: Invalid argument

Ticket: CFE-3807
Changelog: title
(cherry picked from commit 43434600f675b352b1fcbd41ef3054df3270c397)


craigcomstock:
This merge is really for us internally, in our CI where we build DEBUG enabled builds. So we can catch the infrequent lmdb txn invalid argument issue mentioned in CFE-3807.

skaraborgfakir:
I do see this assert in my system. 
Particularly:   in my debian_bookworm physical machine, not in my debian_9 virtual machines.  cfengine-nova 3.18.3

skaraborgfakir:
stopped cfengine (systemctl stop cfengine3)
and 
removed the cf_lock.lmdb - no more pesky messages......

skaraborgfakir:
I got such a message now but on the client which has debian bookworm,  not the other clients which still is debian stretch (they are virtual machines with basically fixed uses.)

PR #1140, opened 2022-10-28
Ensure postgresql is not running before moving its files, not after

Lex-2008:
Ticket: ENT-9547

### DIFF
--- a/packaging/common/cfengine-hub/preinstall.sh
+++ b/packaging/common/cfengine-hub/preinstall.sh
@@ -129,6 +129,77 @@ if is_upgrade; then
   fi
 fi
 
+filter_netstat_listen()
+{
+  set +e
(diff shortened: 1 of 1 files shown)


vpodzime (approved):
ACK. But let's create a ticket to refactor this file (and the even worse postinstal scriptlet) to have the following format:
```
function step1() {...}
function step2() {...}
...

step1
if step2-needed; then
  step2
fi
step3
...
```
Otherwise we will always be facing issues like this.

Lex-2008:
Good point! https://tracker.mender.io/browse/ENT-9555 created.

Lex-2008:
cherry-picks: https://github.com/cfengine/buildscripts/pull/1141, https://github.com/cfengine/buildscripts/pull/1142

PR #1160, opened 2022-11-21
ENT-9600: Added note about why SSLCompression is not explicitly off

nickanderson:
merge together:
-  https://github.com/cfengine/masterfiles/pull/2545
 
When this was removed from buildscripts it also needed to be removed from the
template in masterfiles. We spent some time hunting down the issue before we
realized this was a change we made, not an upstream difference. Rather than
removing it from the config in masterfiles completely we stub it out with an
explanation.

Ticket: ENT-9600
Changelog: None

### DIFF
--- a/deps-packaging/apache/httpd.conf
+++ b/deps-packaging/apache/httpd.conf
@@ -152,6 +152,14 @@ LogLevel warn
   SSLUseStapling off
   # SSLStaplingCache "shmcb:logs/stabling-cache(150000)"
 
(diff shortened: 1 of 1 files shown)


nickanderson:
Per a random STIG https://www.stigviewer.com/stig/apache_server_2.4_windows_server/2019-09-30/finding/V-92469 that says absense of the config is not a finding, I think it's OK that we aren't trying to be smart about the CFEngine version to explicitly disable it.

Lex-2008 (approved):

PR #1192, opened 2023-01-18
Make gzip a weak dependency

Lex-2008:
Investigation shows that the only place where it's _required_ was when compressing man pages. But `man` can also work with uncompressed pages.

Other places which use gzip are Federated Reporting, watchdog script, and self-upgrade on Solaris, - all these parts are not "essential".

So it was decided to downgrade gzip requirement to "Recommends".

Also worth noting that it's essential part of system on many distibutions, and it's pretty hard to find a platform without gzip.

Ticket: ENT-9113

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -32,7 +32,12 @@ fi
 if [ -x /bin/systemctl ]; then
   # This is important in case any of the units have been replaced by the package
   # and we call them in the postinstall script.
-  /bin/systemctl daemon-reload
+  if ! /bin/systemctl daemon-reload; then
+    cf_console echo "warning! /bin/systemctl daemon-reload failed."
+    cf_console echo "systemd seems to be installed, but not working."
+    cf_console echo "Relevant parts of CFEngine installation will fail."
+    cf_console echo "Please fix systemd or use other ways to start CFEngine."
+  fi
 fi
 
 #
@@ -274,7 +279,9 @@ do
   if [ -f /usr/share/man/man8/$i.8.gz ]; then
     rm -f /usr/share/man/man8/$i.8.gz
   fi
-  $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8 && gzip /usr/share/man/man8/$i.8
+  if $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8; then
+    gzip /usr/share/man/man8/$i.8 || true
+  fi
 done
 
 #
--- a/packaging/common/cfengine-non-hub/postinstall.sh
+++ b/packaging/common/cfengine-non-hub/postinstall.sh
@@ -1,7 +1,12 @@
 if [ -x /bin/systemctl ]; then
   # This is important in case any of the units have been replaced by the package
   # and we call them in the postinstall script.
-  /bin/systemctl daemon-reload
+  if ! /bin/systemctl daemon-reload; then
+    cf_console echo "warning! /bin/systemctl daemon-reload failed."
+    cf_console echo "systemd seems to be installed, but not working."
+    cf_console echo "Relevant parts of CFEngine installation will fail."
+    cf_console echo "Please fix systemd or use other ways to start CFEngine."
+  fi
 fi
 
 #
@@ -51,7 +56,9 @@ do
       if [ -f /usr/share/man/man8/$i.8.gz ]; then
         rm -f /usr/share/man/man8/$i.8.gz
       fi
-      $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8 && gzip /usr/share/man/man8/$i.8
+      if $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8; then
+        gzip /usr/share/man/man8/$i.8 || true
+      fi
       ;;
   esac
 done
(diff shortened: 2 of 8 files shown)


packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -274,7 +274,7 @@ do
   if [ -f /usr/share/man/man8/$i.8.gz ]; then
     rm -f /usr/share/man/man8/$i.8.gz
   fi
-  $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8 && gzip /usr/share/man/man8/$i.8
+  $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8 && gzip /usr/share/man/man8/$i.8 || true
```
nickanderson, line 277: Humm, what about || echo "Install gzip and run `$i -M > /usr/share/man/man8/$i.8`

nickanderson (dismissed):
Makes sense, tho I suggested a potential improvement for failed man page (I guess I dunno if the format I suggested would actually work.

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: I don't think it's necessary. On the only platform I've found recently to not contain `gzip`, `man` was accepting uncompressed pages no problem (although other man pages were shipped compressed, inside package) (also worth noting that `man` wasn't installed by default - I had to install it myself).

craigcomstock: You could split this up and check that gzip is available.
```suggestion
  $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8
  command -v gzip 2>/dev/null && gzip /usr/share/man/man8/$i.8
```

Might make a bit more sense pipeline wise. I know shellcheck doesn't like you using `&& ||` due to slightly undefined/imprecise behavior like if `$PREFIX/bin/$i` fails then `true` is still called which isn't really what we want right?

packaging/common/cfengine-non-hub/postinstall.sh
```diff
@@ -51,7 +51,7 @@ do
       if [ -f /usr/share/man/man8/$i.8.gz ]; then
         rm -f /usr/share/man/man8/$i.8.gz
       fi
-      $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8 && gzip /usr/share/man/man8/$i.8
+      $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8 && gzip /usr/share/man/man8/$i.8 || true
```
craigcomstock, line 54: ditto above comment, I think separating this and checking for `gzip` would be better.

craigcomstock (changes requested):

packaging/common/cfengine-hub/postinstall.sh
Lex-2008: > if `$PREFIX/bin/$i` fails then `true` is still called which isn't really what we want right?

Hmm, old code was calling `gzip` only if `$PREFIX/bin/$i -M` succeeded. Maybe we want to preserve this behaviour, too?

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -32,7 +32,12 @@ fi
 if [ -x /bin/systemctl ]; then
   # This is important in case any of the units have been replaced by the package
   # and we call them in the postinstall script.
-  /bin/systemctl daemon-reload
+  if ! /bin/systemctl daemon-reload; then
+    cf_console echo "warning! /bin/systemctl daemon-reload failed."
+    cf_console echo "SystemD seems to be installed, but not working."
```
craigcomstock, line 37:
```suggestion
    cf_console echo "systemd seems to be installed, but not working."
```

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -32,7 +32,12 @@ fi
...
+    cf_console echo "Relevant parts of CFEngine installation will fail."
+    cf_console echo "Please fix SystemD or use other ways to start CFEngine."
```
craigcomstock, line 39:
```suggestion
    cf_console echo "Please fix systemd or use other ways to start CFEngine."
```

packaging/common/cfengine-non-hub/postinstall.sh
```diff
@@ -1,7 +1,12 @@
 if [ -x /bin/systemctl ]; then
   # This is important in case any of the units have been replaced by the package
   # and we call them in the postinstall script.
-  /bin/systemctl daemon-reload
+  if ! /bin/systemctl daemon-reload; then
+    cf_console echo "warning! /bin/systemctl daemon-reload failed."
+    cf_console echo "SystemD seems to be installed, but not working."
```
craigcomstock, line 6:
```suggestion
    cf_console echo "systemd seems to be installed, but not working."
```

packaging/common/cfengine-non-hub/postinstall.sh
```diff
@@ -1,7 +1,12 @@
...
+    cf_console echo "Relevant parts of CFEngine installation will fail."
+    cf_console echo "Please fix SystemD or use other ways to start CFEngine."
```
craigcomstock, line 8:
```suggestion
    cf_console echo "Please fix systemd or use other ways to start CFEngine."
```

craigcomstock (changes requested):

craigcomstock (approved):

PR #1194, opened 2023-01-24
limit "Recommends" to RHEL8+

Lex-2008:


craigcomstock (changes requested):
I think you missed the fix in community spec here: https://github.com/cfengine/buildscripts/blob/master/packaging/cfengine-community/cfengine-community.spec.in#L15

craigcomstock (approved):
молодец!

PR #1246, opened 2023-08-01
ENT-10450: Added docker-images.sh script to build and push hub and agent docker …

craigcomstock:
…images

This script is intended to be used in nightly CI to keep a public repository of docker images up to date.

Ticket: ENT-10450
Changelog: none

### DIFF
--- a/build-scripts/docker-images.sh
+++ b/build-scripts/docker-images.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+set -ex
+registry=${2:-localhost:5001} # TODO make public cfengine AWS ECR registry :)
+
+cat <<EOF >Dockerfile-ubuntu-20-systemd
+FROM ubuntu:20.04
+RUN apt-get update && apt-get install -q -y systemd wget sudo
+CMD [ "/lib/systemd/systemd" ]
+EOF
+
+for name in cfengine-hub cfengine-agent; do
+  if [ "$name" = "cfengine-agent" ]; then
+    install_type="agent"
+  else
+    install_type="hub"
+  fi
+  if docker ps -a | grep $name; then
+    docker ps -a | grep $name | awk '{print $1}' | xargs docker stop
+    docker ps -a | grep $name | awk '{print $1}' | xargs docker rm
+    docker rm $name
+  fi
+  if docker images | grep $name; then
+    docker images | grep $name | awk '{print $3}' | xargs docker rmi --force
+  fi
+  
+  docker build -t $name -f Dockerfile-ubuntu-20-systemd .
+  docker run -d --privileged --name $name $name
+  docker exec -i $name bash -c "wget https://s3.amazonaws.com/cfengine.packages/quick-install-cfengine-enterprise.sh  && sudo bash ./quick-install-cfengine-enterprise.sh $install_type"
+  docker commit $name $name:latest
+  docker image tag $name:latest $registry/cfengine/$name:latest
+  docker image push $registry/cfengine/$name:latest
+  docker stop $name
+  docker rm $name
+  docker rmi $name
+done


build-scripts/docker-images.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+set -ex
+registry=${2:-localhost:5001} # TODO make public cfengine AWS ECR registry :)
```
craigcomstock, line 3: TODO create a public cfengine registry/repository to push to.
TODO provide secrets via jenkins to authenticate to push

build-scripts/docker-images.sh
```diff
@@ -0,0 +1,35 @@
...
+
+cat <<EOF >Dockerfile-ubuntu-20-systemd
```
craigcomstock, line 5: I have found that ubuntu 20 still works with this type of systemd setup. Ubuntu 22 did not.

build-scripts/docker-images.sh
```diff
@@ -0,0 +1,35 @@
...
+    docker ps -a | grep $name | awk '{print $1}' | xargs docker rm
+    docker rm $name
+  fi
+  if docker images | grep $name; then
+    docker images | grep $name | awk '{print $3}' | xargs docker rmi --force
+  fi
+  
+  docker build -t $name -f Dockerfile-ubuntu-20-systemd .
+  docker run -d --privileged --name $name $name
+  docker exec -i $name bash -c "wget https://s3.amazonaws.com/cfengine.packages/quick-install-cfengine-enterprise.sh  && sudo bash ./quick-install-cfengine-enterprise.sh $install_type"
```
craigcomstock, line 28: TODO: replace using the quick install script with using a downloaded package from the `$JOB`.

```wget --no-directories --recursive --no-parent -Acfengine*ubuntu20*deb http://buildcache.cfengine.com/packages/testing-pr/$JOB/```

File: build-scripts/docker-images.sh [hunk 4]
```diff
@@ -0,0 +1,35 @@
...
+  docker commit $name $name:latest
+  docker image tag $name:latest $registry/cfengine/$name:latest
+  docker image push $registry/cfengine/$name:latest
+  docker stop $name
```
craigcomstock, line 32: some cleanup, maybe not quite enough, maybe refactor the cleanup at the beginning of the script.

build-scripts/docker-images.sh
```diff
@@ -0,0 +1,35 @@
...
+  if [ "$name" = "cfengine-agent" ]; then
+    install_type="agent"
+  else
+    install_type="hub"
+  fi
+  if docker ps -a | grep $name; then
+    docker ps -a | grep $name | awk '{print $1}' | xargs docker stop
+    docker ps -a | grep $name | awk '{print $1}' | xargs docker rm
+    docker rm $name
+  fi
```
Lex-2008, line 21: i wonder if you can use
```
docker ps -a --quiet --filter name=$name
```
to get container SHAs, and then pipe them to
```
xargs --no-run-if-empty docker stop
```

Actually, given that `docker stop` prints back SHA of stopped container, could you pipe it to `xargs --no-run-if-empty docker stop`?

```suggestion
  docker ps -a --quiet --filter name=$name | xargs --no-run-if-empty docker stop | xargs --no-run-if-empty docker rm -f
```

build-scripts/docker-images.sh
Lex-2008: actually, we do have something similar for deployment tests [here][x]

[x]: https://github.com/cfengine/system-testing/blob/master/deployment_tests/packagesmapping.py#L154

Lex-2008 (approved):

build-scripts/docker-images.sh
craigcomstock: mm hmm, right, that is a python way of formulating the CORRECT URL instead of globbing through everything and filtering as I suggest above. Cool. No python here. (yet) ;)

build-scripts/docker-images.sh
```diff
@@ -0,0 +1,35 @@
...
+RUN apt-get update && apt-get install -q -y systemd wget sudo
+CMD [ "/lib/systemd/systemd" ]
+EOF
+
+for name in cfengine-hub cfengine-agent; do
+  if [ "$name" = "cfengine-agent" ]; then
+    install_type="agent"
+  else
+    install_type="hub"
+  fi
```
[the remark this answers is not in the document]
craigcomstock: Could. I noticed that beg for refactor as well. ;)

build-scripts/docker-images.sh
craigcomstock: I tend to lean against more arguments and try to keep things "simple" but it is a lovely suggestion.

Lex-2008: yep, that's a valid point: we need to keep a proper balance between "simple" code which is easy to understand by a human, and a code which fits into one line but takes more time to read (by human) and understand. :balance_scale:

PR #1294, opened 2023-09-25
ENT-10647: Added safety for hub upgrade in packaging scriptlets

craigcomstock:
- Improved resilience of hub package installer in case of postgresql server errors
- Added better pg_upgrade failure handling in package install scripts
- Added safety around stopping postgresql server during upgrade

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -437,6 +437,12 @@ init_postgres_dir()
           # Started successfully, stop it again, the migration requires it to be not running.
           (cd /tmp && su cfpostgres -c "$PREFIX/bin/pg_ctl -w -D $PREFIX/state/pg/data -l /var/log/postgresql.log stop")
 
+          # Wait a while if we have to for the server to be stopped
+          if ! wait_for_cf_postgres_down; then
+            cf_console echo "Error: unable to shutdown postgresql server. Showing last of /var/log/postgresql.log for clues."
+            cf_console tail /var/log/postgresql.log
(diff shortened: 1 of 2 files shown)


packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -429,25 +429,37 @@ init_postgres_dir()
...
+        (cd /tmp && su cfpostgres -c "$PREFIX/bin/pg_ctl -w -D $PREFIX/state/pg/data -l /var/log/postgresql.log start")
+        if [ $? = 0 ]; then
+          # Started successfully, stop it again, the migration requires it to be not running.
+          (cd /tmp && su cfpostgres -c "$PREFIX/bin/pg_ctl -w -D $PREFIX/state/pg/data -l /var/log/postgresql.log stop")
+
+          # Wait a while if we have to for the server to be stopped
+          if ! wait_for_cf_postgres_down; then
+            cf_console echo "Error: unable to shutdown postgresql server. Showing last of /var/log/postgresql.log for clues."
+            cf_console tail /var/log/postgresql.log
+            exit 1
```
craigcomstock, line 444: I am a bit worried about this exit here. What state are we leaving things in at this point? I will look at this deeper now.

craigcomstock: I think this is fine, the exit will only exit the subshell and fall through to the code at this message, which is appropriate:

```
cf_console echo "Migration failed. Backup is saved in $BACKUP_DIR."
```

craigcomstock:
need to use non-fork branch :p

PR #1299, opened 2023-09-27
ENT-10647: Added better handling of postgresql server state during hub package install (3.18)

craigcomstock:
The pg_ctl stop/start commands don't ensure that the server state is as reqeusted
at the moment that the command returns. Added up to 5 seconds of wait time to
ensure the state is as we want.

Also added some tails of expected logs when failures occur. Either /var/log/postgresql.log
or the specific pg_upgrade logs which may be created and mentioned in the output of that command
failing.

Ticket: ENT-10647
Changelog: title
(cherry picked from commit a7e6fa707772a9a8e9ca4f4ca307fb4a9a7a2ef2)


craigcomstock:
jenkins looks good, only flakes

nickanderson (approved):

PR #1312, opened 2023-10-05
ENT-10739: Force default 'C' locale for package scriptlets

craigcomstock:
It was found that an LC_CTYPE=UTF-8 from iTerm2 on OSX was breaking the hub package install.

Also use su - (--login) option to preserve this LC_ALL setting

PostgreSQL docs suggest using 'C' locale unless otherwise needed
https://www.postgresql.org/docs/current/locale.html

> The drawback of using locales other than C or POSIX in PostgreSQL is its performance impact. It slows character handling and prevents ordinary indexes from being used by LIKE. For this reason use locales only if you actually need them.

Ticket: ENT-10739
Changelog: title

together
https://github.com/cfengine/system-testing/pull/508 (rhel-8 hub deployment test use 2G instead of 1G RAM)

### DIFF
--- a/packaging/common/script-templates/script-common.sh
+++ b/packaging/common/script-templates/script-common.sh
@@ -1,18 +1,5 @@
-# These can have bad effects on certain commands if set incorrectly. Found
-# when investigating why PostgreSQL was not starting.
-unset LANG
-unset LC_ADDRESS
-unset LC_ALL
-unset LC_IDENTIFICATION
-unset LC_MEASUREMENT
-unset LC_MONETARY
-unset LC_NAME
-unset LC_NUMERIC
-unset LC_PAPER
-unset LC_TELEPHONE
-unset LC_TIME
-
-
+# postgresql docs, https://www.postgresql.org/docs/current/locale.html, recommend using C locale unless otherwise needed
+export LC_ALL=C # overrides all other env vars: https://www.gnu.org/software/libc/manual/html_node/Locale-Categories.html
 # Upgrade detection is a mess. It is often difficult to tell, especially from
 # the postinstall script, so we use the package-upgrade.txt file to remember.
 case "$PKG_TYPE" in


craigcomstock:
Tested this manually/locally with iTerm2 (but any shell will do really) when I set `LC_CTYPE=UTF-8` and install with `cf-remote install --hub vagrant@ubuntu-20` (aka versinon 3.21.2) I get the error in the ticket.

When I keep that `LC_CTYPE` env var and install the package from this ticket available in the github action artifacts: https://github.com/cfengine/buildscripts/suites/16918978368/artifacts/966586445 the installation succeeds! :)

craigcomstock:
@cf-bottom jenkins please

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9931/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9931/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9931/

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -411,7 +413,7 @@ init_postgres_dir()
   # Note: postgres expects $PWD to be writeable, so all postgres commands
   # should be executed from cfpostgres-writeable directory.
   # /tmp is such directory on most cases
-  (cd /tmp && su cfpostgres -c "$PREFIX/bin/initdb -D $PREFIX/state/pg/data")
+  (cd /tmp && su - cfpostgres -c "$PREFIX/bin/initdb -D $PREFIX/state/pg/data")
```
vpodzime, line 416: Doesn't `su -` (I would actually prefer `su -l`) actually **clear** env variables (except for some whitelisted ones)?

vpodzime (commented):
Good job! I'm just not sure about the `su -` change.

packaging/common/cfengine-hub/postinstall.sh
craigcomstock: ah, so strange... I read the man page incorrectly. So I would suggest that my use of `su -` actually fixes the problem by removing `LC_CTYPE` from the env before running. Yes, `-`, `-l`, and `--login` all seem to be the same according to linux su man page. Which in some ways is safer so we could just change `su -` to `su --login` to be clear and remove the `LC_ALL=C` in both cases.

craigcomstock: Made the change, will test the github generated package manually and then try jenkins.

craigcomstock:
looks like oslo-dc shutdown during the last jenkins run.

```
12:00:20 ./27_cf-secret/encrypt-decrypt.cf Pass
12:00:20 ./27_cf-secret/encrypt-decrypt-args.cf Pass
12:00:20 FATAL: command execution failed
12:07:47 java.io.EOFException
12:07:47 	at java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2799)
12:07:47 	at java.io.ObjectInputStream$BlockDataInputStream.readShort(ObjectInputStream.java:3274)
```

craigcomstock:
@cf-bottom jenkins please, thanks. just run a smidge ahead of nightlites eh? ;)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9932/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9932/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9932/

packaging/common/cfengine-hub/postinstall.sh
```diff
@@ -411,7 +411,7 @@ init_postgres_dir()
   # Note: postgres expects $PWD to be writeable, so all postgres commands
   # should be executed from cfpostgres-writeable directory.
   # /tmp is such directory on most cases
-  (cd /tmp && su cfpostgres -c "$PREFIX/bin/initdb -D $PREFIX/state/pg/data")
+  (cd /tmp && su --login cfpostgres -c "$PREFIX/bin/initdb -D $PREFIX/state/pg/data")
```
vpodzime, line 414: `cd /tmp` is useless with `su --login` because one of the things it does is:
> changes to the target user’s home directory

Not sure what that home directory is. I'd actually go with setting and exporting `LC_ALL` and avoiding `--login`. The change to use `--login` can bring many unexpected differences.

craigcomstock: mmm, pretty sure cfpostgres HAS no home dir, that's why we use /tmp, I will check.

Interesting. I suppose by default su keeps env vars. I suppose less is more sometimes, e.g. maybe someone forgets `--login` but they can't forget `su cfpostgres` ;)

craigcomstock: yeah, no homedir, so I will adjust

```
[root@rhel9 vagrant]# su --login cfpostgres
Last login: Fri Oct  6 13:40:14 UTC 2023
su: warning: cannot change directory to /home/cfpostgres: No such file or directory
[cfpostgres@rhel9 vagrant]$ pwd
/home/vagrant
```

craigcomstock: actually, there is already scriptlet code to unset MOST locale stuff, was missing `LC_CTYPE` and so my `export LC_ALL` was nullified actually. So am refactoring to include `LC_CTYPE` and `LC_COLLATE`

packaging/common/script-templates/script-common.sh
```diff
@@ -3,6 +3,8 @@
 unset LANG
 unset LC_ADDRESS
 unset LC_ALL
+unset LC_CTYPE
```
craigcomstock, line 6: @vpodzime I tried two strategies: add more env vars here and added `export LC_ALL=C` in the pre/post scriptlets. Essentially the same difference but I prefer the `LC_ALL=C` because it is an "override" that erases the effect of any other `LC_*` or `LANG*` env vars.

https://www.gnu.org/software/libc/manual/html_node/Locale-Categories.html

> LC_ALL
This is not a category; it is only a macro that you can use with setlocale to set a single locale for all purposes. Setting this environment variable overwrites all selections by the other LC_* variables or LANG.

I will test only using LC_ALL and if it works, remove this whole block of unset as they are ineffectual if we have LC_ALL in place. Actually, might just replace this unset block with one `export LC_ALL=C` instead.

vpodzime: Sound promising!

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9939/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9939/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9939/

craigcomstock:
sequential-tests broke in upgrade-triple test on masterfiles update from 3.18.5 to master, apache wasn't stopped when the policy expected it to be. Known issue? Flake? I doubt it has anything to do with this change as I see no errors in scriptlets/package installs.

craigcomstock:
investigating odd new errors in hub upgrades.

```
error: Couldn"t fork %transfiletriggerun(systemd-239-74.el8_8.5.x86_64): Cannot allocate memory
```

craigcomstock:
> investigating odd new errors in hub upgrades.
> 
> ```
> error: Couldn"t fork %transfiletriggerun(systemd-239-74.el8_8.5.x86_64): Cannot allocate memory
> ```

Could not reproduce with a local vagrant/virtual box VM. Will try the same ami/config as jenkins now.

craigcomstock:
retrying with 2GB rhel-8 instead of 1GB. In manual testing I had no problems but have had problems here several times. I can't find any reference to LC_ALL=C causing larger memory usage but... maybe?

[Build Status](https://ci.cfengine.com/job/sequential-tests/6288/)

craigcomstock:
@cf-bottom jenkins

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/9948/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/9948/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-9948/

vpodzime:
> LC_ALL=C causing larger memory usage but... maybe?

hardly

vpodzime (approved):
ACK

packaging/common/script-templates/script-common.sh
```diff
@@ -1,18 +1,5 @@
...
-unset LC_ADDRESS
-unset LC_ALL
-unset LC_IDENTIFICATION
-unset LC_MEASUREMENT
-unset LC_MONETARY
-unset LC_NAME
-unset LC_NUMERIC
-unset LC_PAPER
-unset LC_TELEPHONE
-unset LC_TIME
```
olehermanse, line 13: At first glance it looks safer to leave these. But I guess you've read through the appropriate man pages and concluded that it's safe to trust only setting `LC_ALL`.

olehermanse (approved):

packaging/common/script-templates/script-common.sh
craigcomstock: Yes. It is fairly clear that `LC_ALL` is a "override all" and in practice/testing I have found it so as well.

https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html

> LC_ALL is an environment variable that overrides all of these. It is typically used in scripts that run particular programs. For example, configure scripts generated by GNU autoconf use LC_ALL to make sure that the configuration tests don’t operate in locale dependent ways.

craigcomstock:
cherry picks
https://github.com/cfengine/buildscripts/pull/1321
https://github.com/cfengine/buildscripts/pull/1320

PR #1325, opened 2023-10-24
ENT-10801: Moved sftp cache Wix Tools install from jenkins-vms to package-msi script

craigcomstock:
Might as well move the knowledge as close to where it is needed to avoid trouble.

Ticket: ENT-10801
Changelog: none


vpodzime (commented):
I'm not sure about this. Installation of tools required for the build  has traditionally been done in the build machine setup.

craigcomstock:
> I'm not sure about this. Installation of tools required for the build has traditionally been done in the build machine setup.

Sure. But I would prefer that the knowledge of how to setup a build machine is close to where a build machine is used: here.

Ideally I'd like to see all the distribution package dependencies moved here as well.

e.g. https://github.com/cfengine/buildscripts/blob/master/ci/setup.sh and https://github.com/cfengine/buildscripts/blob/master/ci/setup-projects.sh made more generic so they work on all supported platforms.

vpodzime (approved):
As discussed on Slack this should be a temporary fix until we move mingw builds into containers which should then have all the necessary bits in the container image(s).

craigcomstock:
cherry picks
https://github.com/cfengine/buildscripts/pull/1331
https://github.com/cfengine/buildscripts/pull/1330

PR #1373, opened 2024-01-26
ENT-11166: Ubuntu-20 build host needs psycopg2 for python2

craigcomstock:
Ticket: ENT-11166
Changelog: none


craigcomstock:
used this branch on oslo-dc to restart ubuntu-20 hosts and they look good, have psycopg2 for root with python2, which is what nova/tests/reporting needs and which is equivalent to what was in jenkins-vms previously.

larsewi (approved):

nickanderson (approved):

PR #1384, opened 2024-01-31
ENT-11166: Adjusted rhel-8 and rhel-9 build host setup to use best=False

craigcomstock:
Instead of adding --nobest option to commands, rework the /etc/(yum&dnf).conf files

Ticket: ENT-11166
Changelog: none


craigcomstock:
@nickanderson gave a verbal/video call OK to this quick fix for jenkins upgrade. :)

PR #1394, opened 2024-02-05
ENT-11233: Aborted package installation if backup directory exists non-empty

nickanderson:
We have observed the presence of an old backup directory causing issues during
upgrade. This change avoids proceeding with package installation if a backup
directory exists and is not empty.

Previously this was restricted to when we thought we would be doing a database
migration, but we have seen issues with failed migration when the backup
directory is non-empty and we are not doing any migration, so it is better to
simply avoid installing the package if the backup directory is not empty even if
we don't expect to use it for migration.

Ticket: ENT-11233
Changelog: Title


olehermanse:
Cherry-pick: https://github.com/cfengine/buildscripts/pull/1428

PR #1401, opened 2024-03-27
Use reflinks if possible for copying masterfiles and web UI

vpodzime:
The postinstall scriptlet on hub copies masterfiles and web UI to separate the pristine original version from the working copy ready for customization. By using `--reflink=auto` will tell `cp` to try to use reflinks if possible which makes the copy much faster on file systems with reflinks enabled (XFS, btrfs, bcachefs,...). It also saves disk space, but checking `du -sh /var/cfengine` gives the same result because `du` doesn't inspect the FS into such detail.

Ticket: none
Changelog: Postinstallation scriptlets of hub packages now use
           reflinks to copy masterfiles and web UI files if
           possible

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -64,7 +64,7 @@ if [ ! -f "$PREFIX/ppkeys/localhost.priv" ]; then
 fi
 
 if [ ! -f "$PREFIX/masterfiles/promises.cf" ]; then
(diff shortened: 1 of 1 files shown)


vpodzime:
jenkins: [Build Status](https://ci.cfengine.com/job/pr-pipeline/)

craigcomstock (approved):
seems reasonable. Is --reflink supported widely? I checked MacOS, it is there. I wonder about BSDs/exotics but we don't support hub there currently. :)

vpodzime:
> I wonder about BSDs/exotics but we don't support hub there currently. :)

Exactly. :)

PR #1408, opened 2024-04-30
Debian 10 (buster) repos are aging out due to support ending soon

craigcomstock:
Switch to archive.debian.org

Ticket: none


craigcomstock:
Apparently this is not needed. We are using the debian-10-backports AMI which apparently had a blip yesterday.

https://wiki.debian.org/Cloud/AmazonEC2Image/Buster mentions both plain buster which just got an update April 30 (yesterday) and buster backports which apparently broke for a moment.

PR #1409, opened 2024-04-30
Update quick install script checksum after an official change was made

craigcomstock:
Ticket: none


larsewi (approved):

olehermanse:
@craigcomstock should we add an automation to create PRs like this?

PR #1418, opened 2024-05-16
ENT-11798: Updated openldap-LMDB checksum

craigcomstock:
For some reason the checksum changed even though the source version/tag has not.

I checked manually and the checksum is indeed different.

Ticket: ENT-11798
Changelog: none


vpodzime:
> For some reason the checksum changed even though the source version/tag has not.

It could be just a tarball created with a different version of tar/gzip or there could be timestamps somewhere.

vpodzime (approved):
ACK. I've seen this as well a week ago.

craigcomstock:
I checked and no files changed, only a .gitignore file was added.

craigcomstock:
no need to cherry pick as 3.18.x and 3.21.x already have been updated to 0.9.32

PR #1423, opened 2024-05-23
ENT-11808: detect-environment: Fixed version detection for RHEL

larsewi:
Jenkins failed with "Unknown RHEL Server version: 8.10" because the
`detect-environment` script did not take into account that the minor
version number can have more than one decimal.

Ticket: ENT-11808
Changelog: None

[Build Status](https://ci.cfengine.com/job/pr-pipeline/10801/)


vpodzime (approved):
ACK

PR #1459, opened 2024-06-19
Only use 'hostname -I' to get local IPs for subjectAltName

vpodzime:
The hostname(1) man page says for the '-i' option:

  Avoid using this option; use hostname --all-ip-addresses instead.

where '-I' is the short version of '--all-ip-addresses' for which:

  The loopback interface and IPv6 link-local addresses  are  omitted.

So we don't have to filter those out and worry about the link-local addresses reported like this:

  fe80::cf:85ff:fe1c:54d7%enX0

which is not a valid IP address and OpenSSL fails on it.

Ticket: ENT-7598
Changelog: None


vpodzime:
jenkins (DTs only): [Build Status](https://ci.cfengine.com/job/pr-pipeline/10949/)

craigcomstock (approved):

PR #1475, opened 2024-07-02
ENT-11724: bump cfbs to 4.1.1 to get -M option

craigcomstock:
postinstall scripts assume cfbs has -M option but this only comes with release 4.1.1


mineralsfree (approved):

vpodzime (approved):
ACK

PR #1484, opened 2024-07-24
Update quick-install script sha for build host setup script

craigcomstock:
Ticket: ENT-11746
Changelog: none


nickanderson (approved):
That's correct.

PR #1488, opened 2024-08-07
try to use larger runner arm64-ubuntu-22 to build an arm64 ubuntu-22 hub package

craigcomstock:
Ticket: ENT-12100


craigcomstock:
sure, the build failed due to php-7.4* packages not available in ubu22 as they are in ubu20. I will work through this container build locally and try again after I get it working there.

I can see now that the arm64 larger runner works.

craigcomstock:
stale, don't need.

PR #1509, opened 2024-10-14
Fixed usage of pg_upgrade in hub package upgrades

craigcomstock:
In some cases library changes in libpq.so or other dependencies can change and since pg_upgrade
checks multiple binaries in both old and new bindirs we need to include both old and new lib dirs in LD_LIBRARY_PATH.

See check_bin_dir() in pg_upgrade code: https://github.com/postgres/postgres/blob/master/src/bin/pg_upgrade/exec.c#L383

Ticket: ENT-12383
Changelog: title

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -525,7 +525,7 @@ check_disk_space() {
 #   and then importing it into new one
 
 migrate_db_using_pg_upgrade() {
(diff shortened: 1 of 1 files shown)


craigcomstock:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/11277/)

craigcomstock:
prefer https://github.com/cfengine/buildscripts/pull/1511

PR #1510, opened 2024-10-15
Moved PostgreSQL to WORKDIR/PostgreSQL/<psql-major-version> to help with hub package upgrades migration

craigcomstock:
pg_upgrade will work better if we have a major-version specific directory in which libraries are contained.

Without this we end up with commands like psql linked to /var/cfengine/lib/libpq.so in both old and new versions during upgrades.
libpq.so can easily change in a breaking way and no amount of LD_LIBRARY_PATH can fix pg_upgrade because it checks BOTH old and new bin dirs with the same execution and has not argument to specify a different library dir for old and new versions.

Ticket: ENT-12383
Changelog: title


craigcomstock:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/11285/)

craigcomstock:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/11286/)

craigcomstock:
prefer https://github.com/cfengine/buildscripts/pull/1511

PR #1520, opened 2024-11-07
ENT-12414: Added librsync to deps-packaging

larsewi:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/11510/)


cf-bottom:

larsewi:
@craigcomstock this does not work yet. Any idea what I might be missing or if I did something wrong? To be honest, I don't know what I'm doing :stuck_out_tongue_winking_eye:

larsewi:
> @craigcomstock this does not work yet. Any idea what I might be missing or if I did something wrong? To be honest, I don't know what I'm doing 😜

@craigcomstock I got a little bit further, so you don't need to help me (yet) :wink:

ci/cfengine-build-host-setup.cf
```diff
@@ -53,6 +53,7 @@ bundle agent cfengine_build_host_setup
       "mingw-w64";
     (debian_10|debian_11).systemssl_build_host::
       "libssl-dev";
+      "librsync-dev";
```
craigcomstock, line 56: This looks like a very wrong place for this. I would expect librsync-dev to be under some "any" or more generic debian/redhat class guards.

deps-packaging/librsync/cfbuild-librsync.spec
```diff
@@ -0,0 +1,101 @@
...
+
+./configure --prefix=%{prefix} --enable-shared --disable-static
+
+rm -f ${RPM_BUILD_ROOT}%{prefix}/lib/librsync.a
+rm -f ${RPM_BUILD_ROOT}%{prefix}/lib/librsync.la
+
+%build
+
+if [ -z $MAKE ]; then
+  MAKE_PATH=`which MAKE`
```
craigcomstock, line 61: maybe use `command -v`? I have seen places with no `which` for sure.

Not sure about the command `which MAKE` do you mean `which make`?

craigcomstock (commented):
a couple of questions/comments, looks good!

ci/cfengine-build-host-setup.cf
larsewi: It's only needed for the host running the `bootstrap-tarballs` script. Not sure where to put it

deps-packaging/librsync/cfbuild-librsync.spec
larsewi: I see that other spec files don't use the `$MAKE` environment variable. So I guess I can just call `make` directly.

ci/cfengine-build-host-setup.cf
craigcomstock: There isn't a section currently. We would need to add a flag file like I did for mingw and adjust jenkins-vms as well.

ci/cfengine-build-host-setup.cf
```diff
@@ -144,6 +146,7 @@ bundle agent cfengine_build_host_setup
     any::
       "mingw_build_host" expression => fileexists("/etc/cfengine-mingw-build-host.flag");
       "systemssl_build_host" expression => fileexists("/etc/cfengine-systemssl-build-host.flag");
+      "bootstrap_pr_host" expression => fileexists("/etc/cfengine-bootstrap-pr-host.flag");
```
craigcomstock, line 149: very cool. Thank you for doing this!

Sorry for the trouble but ideally you would ensure that the packages installed in jenkins-vms are replaced with bits in this policy since you are using that now.

https://gitlab.com/Northern.tech/CFEngine/jenkins-vms/-/merge_requests/59#note_2245675854

larsewi: Alright, let me do that in a follow-up PR :smile:

larsewi:
HP-UX failure is probably due to https://github.com/cfengine/buildscripts/pull/1546

craigcomstock (approved):

ci/cfengine-build-host-setup.cf
```diff
@@ -50,6 +50,8 @@ bundle agent cfengine_build_host_setup
       "mingw-w64";
     (debian_10|debian_11).systemssl_build_host::
       "libssl-dev";
+    debian_9.bootstrap_pr_host::
```
vpodzime, line 53: I know you just followed the pattern, but I don't see why this should be specific to `debian_9`. When we, hopefully, upgrade the machine one day, it will still need `librsync-dev`, right? Let's just do `debian` so that we don't try to install it on an RPM-based system with different package naming scheme, but that's all we need, IMHO.

PR #1524, opened 2024-11-12
Install python3-psycopg2 on ubuntu >= 20

craigcomstock:
Nova reporting tests have switched to preferring python3 if available, which it is on ubuntu >= 20.

Ticket: ENT-12432
Changelog: none


larsewi (approved):

craigcomstock:
cherry pick just to stay up-to-date and make other cherry picks easier. master copy is the only one used in CI so far.

PR #1558, opened 2024-12-13
Bumped CI ubuntu from 20 to 22

craigcomstock:
We were getting odd errors with ldap.so for php:

/usr/lib/php/20190902/ldap.so: undefined symbol: RETURN_THROWS

Ticket: ENT-12530
Changelog: none


craigcomstock:
let's make sure that dependent workflows also work: https://github.com/cfengine/mission-portal/pull/2702

nickanderson (approved):

PR #1570, opened 2025-01-16
Fixed issue where rhel >8 packages would not have correct openssl dependency version

craigcomstock:
We build against systems with the latest available dependencies such as OpenSSL.
libcurl builds against this OpenSSL version which can have API changes that will break functionality if used on a non-updated system.

Ticket: ENT-12587
Changelog: title

### DIFF
--- a/build-scripts/package
+++ b/build-scripts/package
@@ -142,11 +142,16 @@ case "$PACKAGING" in
         fi
     fi
 
-    # determine policy-version for rhel systems so we can require what we build with
+    # determine the system-provided versions of dependencies we build against so we can Require them later in our RPM spec files.
     if [ "$OS" = "rhel" ]; then
         SELINUX_POLICY_VERSION=$(rpm -q --qf '%{VERSION}\n' selinux-policy)
         if [ -z "$SELINUX_POLICY_VERSION" ]; then
             echo "error: unable to determine selinux-policy package version"
+            exit 1
+        fi
+        OPENSSL_VERSION=$(rpm -q --provides openssl-libs | grep OPENSSL_ | sed 's/^.*_\([0-9.]*\).*$/\1/' | sort -n | tail -1)
+        if [ -z "$OPENSSL_VERSION" ]; then
+            echo "error: unable to determine openssl package version"
             exit 1
         fi
     fi
@@ -155,6 +160,7 @@ case "$PACKAGING" in
         -e "s/@@VERSION@@/$RPM_VERSION/g" \
         -e "s/@@RELEASE@@/$safe_prefix$RPM_RELEASE/g" \
         -e "s/@@SELINUX_POLICY_VERSION@@/$SELINUX_POLICY_VERSION/g" \
+        -e "s/@@OPENSSL_VERSION@@/$OPENSSL_VERSION/g" \
         -e "/^%pre\$/r $PREINSTALL" \
         -e "/^%post\$/r $POSTINSTALL" \
         -e "/^%preun\$/r $PREREMOVE" \
--- a/packaging/cfengine-nova-hub/cfengine-nova-hub.spec.in
+++ b/packaging/cfengine-nova-hub/cfengine-nova-hub.spec.in
@@ -33,9 +33,13 @@ Requires: libssl.so.1.1()(64bit) libssl.so.1.1(OPENSSL_1_1_0)(64bit) libssl.so.1
 Requires: libcrypto.so.1.1()(64bit) libcrypto.so.1.1(OPENSSL_1_1_0)(64bit)
 %endif
 
+# We build against systems with the latest available dependencies such as OpenSSL.
+# We use rpm -q --provides to determine the highest API present in OpenSSL and then use that as a Requires.
+# OPENSSL_VERSION is determined in build-scripts/package script.
+# This should ensure that when packages are installed with yum/dnf any required OpenSSL package upgrades will be performed or the installation will fail.
 %if %{?rhel}%{!?rhel:0} > 8
-Requires: libcrypto.so.3()(64bit) libcrypto.so.3(OPENSSL_3.0.0)(64bit) libcrypto.so.3(OPENSSL_3.0.1)(64bit)
-Requires: libssl.so.3()(64bit) libssl.so.3(OPENSSL_3.0.0)(64bit)
+Requires: libcrypto.so.3()(64bit) libcrypto.so.3(OPENSSL_@@OPENSSL_VERSION@@)(64bit)
+Requires: libssl.so.3()(64bit) libssl.so.3(OPENSSL_@@OPENSSL_VERSION@@)(64bit)
 %endif
 
 # cfbs/Build requires Python 3.5+ (not available on RHEL 6)
--- a/packaging/cfengine-nova/cfengine-nova.spec.in
+++ b/packaging/cfengine-nova/cfengine-nova.spec.in
@@ -29,9 +29,13 @@ Requires: libssl.so.1.1()(64bit) libssl.so.1.1(OPENSSL_1_1_0)(64bit) libssl.so.1
 Requires: libcrypto.so.1.1()(64bit) libcrypto.so.1.1(OPENSSL_1_1_0)(64bit)
 %endif
 
+# We build against systems with the latest available dependencies such as OpenSSL.
+# We use rpm -q --provides to determine the highest API present in OpenSSL and then use that as a Requires.
+# OPENSSL_VERSION is determined in build-scripts/package script.
+# This should ensure that when packages are installed with yum/dnf any required OpenSSL package upgrades will be performed or the installation will fail.
 %if %{?rhel}%{!?rhel:0} > 8
-Requires: libcrypto.so.3()(64bit) libcrypto.so.3(OPENSSL_3.0.0)(64bit) libcrypto.so.3(OPENSSL_3.0.1)(64bit)
-Requires: libssl.so.3()(64bit) libssl.so.3(OPENSSL_3.0.0)(64bit)
+Requires: libcrypto.so.3()(64bit) libcrypto.so.3(OPENSSL_@@OPENSSL_VERSION@@)(64bit)
+Requires: libssl.so.3()(64bit) libssl.so.3(OPENSSL_@@OPENSSL_VERSION@@)(64bit)
 %endif
 
 AutoReqProv: no


build-scripts/package
```diff
@@ -129,19 +129,25 @@ case "$PACKAGING" in
...
 
-    # determine policy-version for rhel systems so we can require what we build with
+    # determine system-provided dependency versions that we build with so that installs work well
     if [ "$OS" = "rhel" ]; then
         SELINUX_POLICY_VERSION=$(rpm -q --qf '%{VERSION}\n' selinux-policy)
         if [ -z "$SELINUX_POLICY_VERSION" ]; then
             echo "error: unable to determine selinux-policy package version"
             exit 1
         fi
+        OPENSSL_VERSION=$(rpm -q --qf '%{VERSION}\n' openssl | sed  's/\([0-9]*.[0-9]*.\)[0-9]*/\10/')
```
vpodzime, line 139: It's not always `x.y.0`, see this `rpm -q --provides openssl-libs` output on my system:
```
libcrypto.so.3()(64bit)
libcrypto.so.3(OPENSSL_3.0.0)(64bit)
libcrypto.so.3(OPENSSL_3.0.1)(64bit)
libcrypto.so.3(OPENSSL_3.0.3)(64bit)
libcrypto.so.3(OPENSSL_3.0.8)(64bit)
libcrypto.so.3(OPENSSL_3.0.9)(64bit)
libcrypto.so.3(OPENSSL_3.1.0)(64bit)
libcrypto.so.3(OPENSSL_3.2.0)(64bit)
libssl.so.3()(64bit)
libssl.so.3(OPENSSL_3.0.0)(64bit)
libssl.so.3(OPENSSL_3.2.0)(64bit)
openssl-libs = 1:3.2.2-3.fc40
openssl-libs(x86-64) = 1:3.2.2-3.fc40
config(openssl-libs) = 1:3.2.2-3.fc40
libcrypto.so.3
libcrypto.so.3(OPENSSL_3.0.0)
libcrypto.so.3(OPENSSL_3.0.1)
libcrypto.so.3(OPENSSL_3.0.3)
libcrypto.so.3(OPENSSL_3.0.8)
libcrypto.so.3(OPENSSL_3.0.9)
libcrypto.so.3(OPENSSL_3.1.0)
libcrypto.so.3(OPENSSL_3.2.0)
libssl.so.3
libssl.so.3(OPENSSL_3.0.0)
libssl.so.3(OPENSSL_3.2.0)
openssl-libs = 1:3.2.2-3.fc40
openssl-libs(x86-32) = 1:3.2.2-3.fc40
```

We should take the highest version from the above, I think.

vpodzime (commented):
Looks good to me otherwise. One thing that worries me, though, is that I knew the fixed `3.0.0`/`3.0.1` would stop working at some point, but I expected our tests to catch this. How come that is not the case?

build-scripts/package
craigcomstock: Yeah, I tried 3.2.2 aka latest first but that didn't work well. I'll have to remember/rediscover why.

I don't know that there is one answer that "just works" in this case.

craigcomstock: Probably a better strategy would be to know that libcurl has this dependency on openssl, among other packages, and to examine those binaries (.so files) and find the max symbol in there.

Ideally we would repeat that for other libraries we build against but don't vendor, in case they include references for a specific version of the library that folks could not have updated before installing/upgrading cfengine.

```
[root@rocky9 lib]# objdump -x libpromises.so.3 | grep " OPENSSL_[0-9]"
    0x06702b20 0x00 04 OPENSSL_3.0.0
    0x06702b20 0x00 03 OPENSSL_3.0.0
[root@rocky9 lib]# objdump -x libcurl.so.4 | grep " OPENSSL_[0-9]"
    0x06702b20 0x00 04 OPENSSL_3.0.0
    0x06702d20 0x00 11 OPENSSL_3.2.0
    0x06702b20 0x00 03 OPENSSL_3.0.0
[root@rocky9 lib]# objdump -x libldap.so.2 | grep " OPENSSL_[0-9]"
    0x06702b20 0x00 08 OPENSSL_3.0.0
    0x06702b20 0x00 03 OPENSSL_3.0.0
```

e.g. libpromises.so has these dependencies, not all of which we vendor. Really just openssl, glibc and acl.

```
Version References:
  required from libm.so.6:
    0x06969189 0x00 13 GLIBC_2.29
    0x09691a75 0x00 09 GLIBC_2.2.5
  required from libacl.so.1:
    0x05822450 0x00 07 ACL_1.0
  required from libssl.so.3:
    0x06702b20 0x00 04 OPENSSL_3.0.0
  required from libcrypto.so.3:
    0x06702b20 0x00 03 OPENSSL_3.0.0
  required from libc.so.6:
    0x06969187 0x00 17 GLIBC_2.27
    0x0d696917 0x00 16 GLIBC_2.7
    0x06969197 0x00 15 GLIBC_2.17
    0x06969194 0x00 14 GLIBC_2.14
    0x069691b3 0x00 12 GLIBC_2.33
    0x069691b2 0x00 11 GLIBC_2.32
    0x0d696914 0x00 10 GLIBC_2.4
    0x0d696913 0x00 08 GLIBC_2.3
    0x069691b4 0x00 06 GLIBC_2.34
    0x09691a75 0x00 05 GLIBC_2.2.5
    0x09691972 0x00 02 GLIBC_2.3.2
```

libldap.so has this unique entry:
```
  required from liblber.so.2:
    0x02881810 0x00 05 OPENLDAP_2.200
```

vpodzime: Well, yeah, `objdump -x` is a mess. That's why I suggested we should query the installed `openssl-libs` package we built with and linked against. It's not perfect, but as good as we can go with a relatively simple approach. AFAIK, we don't want to accidentally rely on other things on the system except for OpenSSL for which we do it explicitly and on purpose.

craigcomstock: So you mean something like
```
rpm -q --provides openssl-libs | grep OPENSSL_ | cut -d'(' -f2 | cut -d')' -f1 | cut -d_ -f2 | sort | tail -1
```

or
```
rpm -q --provides openssl-libs | grep OPENSSL_ | sed 's/^.*_\([0-9.]*\).*$/\1/' | sort | tail -1
```

which on my rocky-9 un-upgraded system takes this initial output
```
[vagrant@rocky9 ~]$ rpm -q --provides openssl-libs
config(openssl-libs) = 1:3.0.7-24.el9
libcrypto.so.3()(64bit)
libcrypto.so.3(OPENSSL_3.0.0)(64bit)
libcrypto.so.3(OPENSSL_3.0.1)(64bit)
libcrypto.so.3(OPENSSL_3.0.3)(64bit)
libssl.so.3()(64bit)
libssl.so.3(OPENSSL_3.0.0)(64bit)
openssl-libs = 1:3.0.7-24.el9
openssl-libs(x86-64) = 1:3.0.7-24.el9
```

and results in 3.0.3.

vpodzime: With `sort -n`, but yes. I think that's the best possible easy way.

craigcomstock: yep.

vpodzime (approved):
Not sure if it's worth complicating the commit message, but this change means we require some API level, not a specific version of openssl. The commit message says *hard dependency on the exact version* and that's not the case. Any version that provides the required API level is fine (i.e. basically also any newer version). I'd reword the commit message so that in case someone goes searching for this commit, they don't get the impression that if there's a CVE in OpenSSL, they will be stuck on the some version because of CFEngine.

craigcomstock:
@cf-bottom jenkins with exotics no tests (I will test manually installing on non-updated rhel-8/9)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12092/)

(with exotics) [NO TESTS]

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12092/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12092/

larsewi (approved):

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/12093/)

(with exotics) [NO TESTS]

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/12093/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-12093/

vpodzime:
> They will be required at a minimum VERSION of OpenSSL nothing to do with an API requirement except that yes, the version we build with is the API version we build against.

#### TL;DR

It has to do with API level requirements, really. 

#### Just to elaborate on what I meant

The *provides* returned by `rpm -q --provides` are based on symbol names/versions in the libraries' ELF files. So even if we build with `openssl-3.5.0` the highest symbol name/version in the `libssl` and `libcrypto` libraries will determine the provides and thus our requires. If we wanted to go with a finer-grained approach, we would check the libraries and binaries we build and determine the highest symbol/name version based on what we actually link with. Then, even when building with `openssl-3.5.0` we could easily require `3.2.0 or higher` or something similar as long as we and our dependencies didn't use any newer symbols.

Thanks for adjusting the commit message!

vpodzime:
> As I understand it from you this `-q --provides openssl-libs` examines symbols in the libraries and returns

Not exactly. When you do `-q --provides`, it only gives you whatever is stored in the RPM DB (or an RPM package, if you point it to a file). The symbol examination happens when an RPM is built and populates the metadata. Similarly, when an RPM is built, dependencies on dynamic libraries are determined based on the ELF files in the package and their `NEEDED` info. Which is what we disable because we ship our dependencies ourselves.

craigcomstock:
> > As I understand it from you this `-q --provides openssl-libs` examines symbols in the libraries and returns
> 
> Not exactly. When you do `-q --provides`, it only gives you whatever is stored in the RPM DB (or an RPM package, if you point it to a file). The symbol examination happens when an RPM is built and populates the metadata. Similarly, when an RPM is built, dependencies on dynamic libraries are determined based on the ELF files in the package and their `NEEDED` info. Which is what we disable because we ship our dependencies ourselves.

Nice. Thanks for the clarification.

craigcomstock:
Looks good to me:

```
[vagrant@rhel9 ~]$ sudo yum install cfengine-nova-hub-3.26.0a.20fcd8b8c-1692.el9.x86_64.rpm
Extra Packages for Enterprise Linux 9 - x86_64                                                         13 MB/s |  23 MB     00:01    
Extra Packages for Enterprise Linux 9 openh264 (From Cisco) - x86_64                                  794  B/s | 2.5 kB     00:03    
Last metadata expiration check: 0:00:01 ago on Tue 06 May 2025 12:43:40 PM UTC.
Error: 
 Problem: conflicting requests
  - nothing provides selinux-policy >= 38.1.45 needed by cfengine-nova-hub-3.26.0a.20fcd8b8c-1692.el9.x86_64 from @commandline
  - nothing provides libssl.so.3(OPENSSL_3.2.0)(64bit) needed by cfengine-nova-hub-3.26.0a.20fcd8b8c-1692.el9.x86_64 from @commandline
  - nothing provides libcrypto.so.3(OPENSSL_3.2.0)(64bit) needed by cfengine-nova-hub-3.26.0a.20fcd8b8c-1692.el9.x86_64 from @commandline
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

and after upgrading packages the package installs as normal.

vpodzime:
> and after upgrading packages the package installs as normal.

I don't understand why this is necessary. I believe `yum install ...` should pull in the upgrade. Anyway, it works well **from our side**, the behavior above just looks like a bug in yum (dnf, really) to me.

PR #1575, opened 2025-02-04
README.md: Switched to HTTPS urls in dependency tables

olehermanse:
- **README.md: Switched to HTTPS urls in dependency tables**
- **README.md: Replaced broken links**


larsewi (approved):

olehermanse:
For LTS branches, we have intentionally removed these tables to avoid having to keep them up to date. So no need to backport.

PR #1583, opened 2025-02-06
Change unused rsync.log path to specific /tmp/rsync.log (3.24)

craigcomstock:
An instance occurred where maybe the PWD was removed by another process and so an rsync command failed due to not being able to write the file in PWD.

Ticket: ENT-12633
Changelog: none
(cherry picked from commit 3e5e3885b63656244606118ca919b5af9218f02d)

### DIFF
--- a/build-remote
+++ b/build-remote
@@ -183,7 +183,7 @@ prepare_workdir() {
 
 checkout() {
   mkdir -p build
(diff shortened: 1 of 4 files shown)


cf-bottom:
Thank you for submitting a PR! Maybe @larsewi can review this?

PR #1585, opened 2025-02-12
Adjusted cfengine build host policy to use http for debian-9

craigcomstock:
apt-transport-https package is not easily installed so use http instead

Fix for jenkins-vms bootstrap vm.

Ticket: ENT-12659
Changelog: none


nickanderson (approved):
So, I guess this could be constrained to when that https transport package is not installed and then when it is installed, the promise can be for https? Humm, dunno how goood packagesmatching() is with standalone single run policy, but i guess commands could work. 

Makes me also think about dicussion for a validation attribtue on files promises (runa command to verify it's ok before final move into place). Guess it could be if => retunrszero( dpkg -l https-transport) or something.

PR #1587, opened 2025-02-14
Added policy to enable core dumps on all build hosts

craigcomstock:
Should work in most places. Mostly focused here on ENT-12669 bootstrap-pr segfaults during npm i in mission-portal.

Ticket: ENT-12669
Changelog: none

### DIFF
--- a/ci/cfengine-build-host-setup.cf
+++ b/ci/cfengine-build-host-setup.cf
@@ -149,6 +149,8 @@ bundle agent cfengine_build_host_setup
       "mingw_build_host" expression => fileexists("/etc/cfengine-mingw-build-host.flag");
       "systemssl_build_host" expression => fileexists("/etc/cfengine-systemssl-build-host.flag");
       "bootstrap_pr_host" expression => fileexists("/etc/cfengine-bootstrap-pr-host.flag");
+    linux::
+      "have_coredumpctl" expression => returnszero("command -v coredumpctl", "useshell");
     debian_9|ubuntu_16|redhat_6|centos_6::
       "have_opt_jdk21" expression => fileexists("/opt/jdk-21.0.1");
     (redhat|centos).!(redhat_6|centos_6|redhat_7|centos_7)::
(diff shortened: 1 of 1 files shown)


craigcomstock:
need to run on a host first, so draft actually.

ci/cfengine-build-host-setup.cf
```diff
@@ -149,6 +149,8 @@ bundle agent cfengine_build_host_setup
       "mingw_build_host" expression => fileexists("/etc/cfengine-mingw-build-host.flag");
       "systemssl_build_host" expression => fileexists("/etc/cfengine-systemssl-build-host.flag");
       "bootstrap_pr_host" expression => fileexists("/etc/cfengine-bootstrap-pr-host.flag");
+    linux::
+      "have_coredumpctl" expression => returnszero("command -v coredumpctl");
```
nickanderson, line 153: I think this needs a second parameter with "useshell".
```suggestion
      "have_coredumpctl" expression => returnszero("command -v coredumpctl", "useshell");
```

ci/cfengine-build-host-setup.cf
```diff
@@ -173,6 +175,9 @@ bundle agent cfengine_build_host_setup
         comment => "note: centos-7 has installed instead of --installed argument, and that works on rhel-8 and rhel-9 so go with the sub-command instead of option";
 
   commands:
+    have_coredumpctl::
+      "sysctl kernel.core_pattern='|/usr/lib/systemd/systemd-coredump %p %u %g %s %t %e'" -> { "ENT-12669" }
```
nickanderson, line 179: Dosn't this need to be contained within a shell for pipe and execution using $PATH support?
```suggestion
      "sysctl kernel.core_pattern='|/usr/lib/systemd/systemd-coredump %p %u %g %s %t %e'" -> { "ENT-12669" }
        contain => in_shell; # Not sure if this body exists in this standalone file
```

nickanderson (changes requested):
I think the policy needs minor tweaks

craigcomstock:
thanks @nickanderson yes, you are right on both counts, need useshell, will adjust!

craigcomstock:
fixed. tested.

nickanderson (approved):
Seems reasonable to me.

PR #1593, opened 2025-02-20
ENT-12597: Added automatic dependency updates

larsewi:
TODO:
- [x] Ability to skip a specific updates (E.g., `--skip leech 0.1.25 --skip librsync 1-2-3`)
- [ ] ~~Make `/tmp/update-deps-cache.json` a GitHub cached item~~

Back ported to https://github.com/cfengine/buildscripts/pull/1597

### DIFF
--- a/.github/workflows/update-deps.py
+++ b/.github/workflows/update-deps.py
@@ -7,6 +7,7 @@ import argparse
 import requests
 import urllib.request
 import logging as log
+from itertools import batched
 
 DEPS_PACKAGING = "deps-packaging"
 
@@ -24,6 +25,15 @@ def parse_args():
         choices=["major", "minor", "patch"],
         help="whether to do major, minor or patch updates",
     )
+    parser.add_argument(
+        "--skip",
+        nargs=2,
+        action="extend",
+        default=[],
+        metavar=("PACKAGE", "VERSION"),
+        help="skip updates for specific version of a package (e.g., --skip librsync 2.3.4)"
+    )
+
     return parser.parse_args()
 
 
@@ -81,13 +91,27 @@ def get_available_versions(proj_id):
 
 
 def select_new_version(
+    package_name,
     update_type,
+    skip_versions,
     old_version,
     available_versions,
 ):
+    assert len(skip_versions) % 2 == 0 # Is guaranteed by the argument parser
+
     old_split = old_version.replace("-", ".").replace("_", ".").split(".")
     for new_version in available_versions:
         new_split = new_version.replace("-", ".").replace("_", ".").split(".")
+
+        do_skip = False
+        for skip_package, skip_version in batched(skip_versions, 2):
+            skip_split = skip_version.replace("-", ".").replace("_", ".").split(".")
+            if (skip_package == package_name) and (skip_split == new_split):
+                do_skip = True
+        if do_skip:
+            log.info(f"Skipping version {new_version} for package {package_name}")
+            continue
+
         if update_type == "major":
             return new_version
         if update_type == "minor" and old_split[:1] == new_split[:1]:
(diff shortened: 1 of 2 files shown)


.github/workflows/update-deps.py
```diff
@@ -0,0 +1,198 @@
...
+        return version
+
+    log.error(f"Failed to extract version number from '{filename}'")
+    return None
+
+
+def get_available_versions(proj_id):
+    url = f"https://release-monitoring.org/api/v2/versions/?project_id={proj_id}"
+
+    versions_cache = "/tmp/update-deps-cache.json"
```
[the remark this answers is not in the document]
craigcomstock: probably don't want that week-to-week anyway, but locally during a single run makes sense. :+1:

.github/workflows/update-deps.py
```diff
@@ -0,0 +1,198 @@
...
+            continue
+        log.info(f"Updating {pkg_name} from version {old_version} to {new_version}...")
+
+        update_version_numbers(pkg_name, old_version, new_version)
+        update_distfiles_digest(pkg_name)
+
+        commit_message.append(f"- Updated dependency '{pkg_name}' from version {old_version} to {new_version}\n")
+
+    with open("/tmp/commit-message.txt", "w") as f:
+        f.writelines(commit_message)
```
olehermanse, line 219: Is it better with one big commit or individual commits? (Legitimate question, I have no strong opinion)

.github/workflows/update-deps.py
```diff
@@ -0,0 +1,198 @@
...
+    with open(os.path.join("/tmp", tarball), "rb") as f:
+        sha.update(f.read())
+    new_digest = sha.digest().hex()
+
+    replace_string_in_file(filename, old_digest, new_digest)
+
+
+def main():
+    args = parse_args()
+    loglevel = "DEBUG" if args.debug else "INFO"
```
olehermanse, line 179: I like to try to do only arg extraction in `main()`, and then provide a nicer function to call from python code. In this case, you could do:

```python3
def update_deps(loglevel, update):
    ...

def main():
    args = parse_args()
    update_deps(args.loglevel, args.update)
```

.github/workflows/update-deps.py
```diff
@@ -0,0 +1,198 @@
...
+        "--debug",
+        action="store_true",
+        help="enable debug log messages",
+    )
+    parser.add_argument(
+        "--update",
+        default="minor",
+        choices=["major", "minor", "patch"],
+        help="whether to do major, minor or patch updates",
+    )
```
olehermanse, line 27: I think it would be nice if the path to buildscripts root was an argument, so that it doesn't matter where you run the script. For example:

```
python3 .github/workflows/update-deps.py .
python3 .github/workflows/update-deps.py ../other-buildscripts-checkout/
```

.github/workflows/update-deps.py
larsewi: Maybe multiple commits are better? Then you can drop one if it fails, and add `--skip` to the script until it is resolved? Not sure :thinking: Let's discuss it here https://northerntech.atlassian.net/browse/CFE-4503

.github/workflows/update-deps.py
larsewi: See https://northerntech.atlassian.net/browse/CFE-4504

.github/workflows/update-deps.py
larsewi: See https://northerntech.atlassian.net/browse/CFE-4505

.github/workflows/update-deps.py
```diff
@@ -0,0 +1,198 @@
...
+    else:
+        cache = {}
+
+    now = time.time()
+    one_hour = 3600
+    if (url in cache) and (cache[url]["timestamp"] + one_hour) > now:
+        log.debug(f"Retrieving '{url}' from cache '{versions_cache}'")
+        return cache[url]["response"]
+
+    data = requests.get(url).json()
```
vpodzime, line 74: Not a biggie, but this could be done without `requests` as an extra requirement.

PR #1599, opened 2025-02-26
Don't set $MANPATH

abbeyj:
The man pages aren't built anymore since https://github.com/cfengine/core/pull/715 . The directory /var/cfengine/share/man does not exist as part of the package so it does not make sense to refer to it.

### DIFF
--- a/packaging/cfengine-community/profile.sh
+++ b/packaging/cfengine-community/profile.sh
@@ -1,7 +1,3 @@
 if ! echo ${PATH} | /bin/grep /var/cfengine/bin > /dev/null ; then
     export PATH=$PATH:/var/cfengine/bin
 fi
-
-if ! echo ${MANPATH} | /bin/grep /var/cfengine/share/man > /dev/null ; then
-    export MANPATH=$MANPATH:/var/cfengine/share/man
-fi


craigcomstock:
@abbeyj thanks for the contribution! How did you find this?

abbeyj:
A coworker of mine noticed that their `$MANPATH` was pointing at a nonexistent directory and asked why.  I figured out that it was getting set by /etc/profile.d/cfengine3.sh.  This file is part of the cfengine-community RPM but the man pages are not shipped as part of this RPM, or any other RPM that I can find.  So I'm assuming that when the man pages were dropped, removal of this part of profile.sh was overlooked.  It would be easy to miss as nothing will really be broken by leaving this in place.  It will just waste a few cycles on every login and on every invocation of `man`.

craigcomstock:
> A coworker of mine noticed that their `$MANPATH` was pointing at a nonexistent directory and asked why. I figured out that it was getting set by /etc/profile.d/cfengine3.sh. This file is part of the cfengine-community RPM but the man pages are not shipped as part of this RPM, or any other RPM that I can find. So I'm assuming that when the man pages were dropped, removal of this part of profile.sh was overlooked. It would be easy to miss as nothing will really be broken by leaving this in place. It will just waste a few cycles on every login and on every invocation of `man`.

Very good. Thanks for the explanation. I'll take a look around and try to get this merged.

abbeyj:
I believe this is still relevant and I'd still like to have it merged and not closed.

craigcomstock:
> I believe this is still relevant and I'd still like to have it merged and not closed.

Absolutely. I am looking at this now and will try to resolve it one way or another. Thanks for the contribution!

craigcomstock:
This change looks correct to me. We now install man pages to standard locations and not in /var/cfengine/share/man as part of the post install scriptlets: https://github.com/cfengine/buildscripts/blob/master/packaging/common/cfengine-non-hub/postinstall.sh#L55-L64

```
  case `os_type` in
    redhat|debian)
      if [ -f /usr/share/man/man8/$i.8.gz ]; then
        rm -f /usr/share/man/man8/$i.8.gz
      fi
      if $PREFIX/bin/$i -M > /usr/share/man/man8/$i.8; then
        gzip /usr/share/man/man8/$i.8 || true
      fi
      ;;
  esac
```

Man pages printed out with the `-M` option go back to the time of your referenced commit of doc removals and further back.
I checked a rhel-8 system with 3.27.1 cfengine-community repos package and the man pages work fine and are indeed in /usr/share/man/man8 as expected from the scriptlet above.

craigcomstock (approved):

craigcomstock:
@abbeyj Thanks for the contribution! We will take care of cherry picking this change back to LTS releases if we need to. Be well!

PR #1612, opened 2025-03-10
ENT-12669: Fixed pipe failed error when generating core dumps

larsewi:


ci/cfengine-build-host-setup.cf
```diff
@@ -176,7 +176,7 @@ bundle agent cfengine_build_host_setup
 
   commands:
     have_coredumpctl::
-      "sysctl kernel.core_pattern='|/usr/lib/systemd/systemd-coredump %p %u %g %s %t %e'" -> { "ENT-12669" }
+      "sysctl kernel.core_pattern='|/lib/systemd/systemd-coredump %p %u %g %s %t %e'" -> { "ENT-12669" }
```
craigcomstock, line 179: this may only apply to certain platforms? I have a hard time believing that this didn't work "anywhere" as I have often been fooled into thinking that there was an "actual" coredump to debug when I was looking at cf-support tarballs from system tests and would see only one coredump: from the cf-support/sleep test.

But yeah, let's fix things up! Maybe we just need some guards around different distributions for different locations of systemd-coredump.

craigcomstock (approved):
so far I see both the old and proposed paths on debian-12. I figure probably debian-9 (bootstrap) only has the proposed path.

Let's go with this and see how it goes. This is for BUILD hosts (and bootstrap) and not related to system testing hosts provisioned by mass.py in system-testing so we should be "good" on this change.

PR #1621, opened 2025-03-11
ENT-12607: Initial implementation of `deptool.py`, automatic updating of dependencies tables

jakub-nt:


cf-bottom:

olehermanse:
@jakub-nt please show some commands / examples of how to run the tool here in the PR, and put some relevant examples into the README.md file.

larsewi (changes requested):
The _deptool.py_ script does not belong in _build-scripts/_ given that it is not a part of the scripts that actually build CFEngine. Consider moving it to _.github/workflows/_.

When I try to run the program, it crashes with a stack trace after executing `self.run_command("fetch", remote, ref)` on line 204.

build-scripts/deptool.py
```diff
@@ -0,0 +1,728 @@
...
+                )
+            if dep not in HOME_URL:
+                log.warning(
+                    "Dependency " + dep + " is missing from the HOME_URL dictionary"
+                )
+
+        deps_name_urllink_mapping = {
+            d: "[" + HUMAN_NAME.get(d, d) + "](" + HOME_URL.get(d, "0.0.0.0") + ")"
+            for d in all_deps
+        }
```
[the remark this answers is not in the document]
jakub-nt: I've changed it to simply write the dependency name without a link if it's not present in `HOME_URL`.

build-scripts/deptool.py
```diff
@@ -0,0 +1,728 @@
...
+
+        for dep in all_deps:
+            if dep not in HUMAN_NAME:
+                log.warning(
+                    "Dependency " + dep + " is missing from the HUMAN_NAME dictionary"
+                )
+            if dep not in HOME_URL:
+                log.warning(
+                    "Dependency " + dep + " is missing from the HOME_URL dictionary"
+                )
```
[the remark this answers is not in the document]
jakub-nt: It's valid to display the original name and not create a link when they are missing, so I think a warning is more appropriate than an error.

build-scripts/deptool.py
```diff
@@ -0,0 +1,728 @@
...
+
+        md_table = dict_2d_as_markdown_table(
+            compared_deps_data,
+            header_cell="CFEngine version",
+            row_name_mapping=deps_name_urllink_mapping,
+        )
+        return md_table
+
+
+def deps_table_as_cdx(deps_table):
```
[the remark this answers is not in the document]
jakub-nt: It's a SBOM format: https://cyclonedx.org/. Added a docstring.

build-scripts/deptool.py
```diff
@@ -0,0 +1,728 @@
+import argparse
+import collections
+import json
+import logging as log
+import os
+import re
+import subprocess
+import sys
+
+ACTIVE_BRANCHES = ["3.21.x", "3.24.x", "master"]
```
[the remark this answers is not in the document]
olehermanse: https://northerntech.atlassian.net/browse/ENT-12709

scripts/deptool.py
```diff
@@ -0,0 +1,728 @@
...
+        print("\n### Enterprise Hub dependencies:\n")
+        print(updated_hub_table)
+
+    if args.patch:
+        dr.patch_readme(updated_readme)
+
+
+if __name__ == "__main__":
+    ret = main()
+    sys.exit(ret)
```
olehermanse, line 728: This technically creates a global variable which could be avoided;

```suggestion
if __name__ == "__main__":
    sys.exit(main())
```

olehermanse: (but this is not a big deal, of course)

build-scripts/deptool.py
```diff
@@ -8,6 +8,70 @@
 import sys
 
 
+HUMAN_NAME = {
```
craigcomstock: I don't see how this human name is much better than the existing non-human name. Can you give an example of something that isn't obvious?

build-scripts/deptool.py
```diff
@@ -8,6 +8,70 @@
...
+    "git": "Git",
+    "libexpat": "libexpat",
+    "php": "PHP",
+    "postgresql": "PostgreSQL",
+    "nghttp2": "nghttp2",
+    "rsync": "rsync",
+    "lcov": "LCOV",
+    "libcurl-hub": "libcurl-hub",
+}
+HOME_URL = {
```
craigcomstock: Adding these URLs here seems like a maintenance problem. Can we retrieve the URL from a single source like the deps/<dep> dir/files?

craigcomstock (commented):
Cool. Looks good. Some comments!

build-scripts/deptool.py
```diff
@@ -0,0 +1,417 @@
...
+            # already added. In this case, we should succeed in setting its url.
+            self.run_command("remote", "set-url", "upstream", upstream_url)
+        if checkout_branch is not None:
+            self.checkout(checkout_branch)
+        if checkout_tag is not None:
+            self.checkout(checkout_tag, tag=True)
+
+    def run_command(self, *command, **kwargs):
+        """Runs a git command against git repo.
+        Syntaxically this function tries to be as close to subprocess.run
```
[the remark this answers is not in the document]
craigcomstock: you fixed this later, ack.

build-scripts/deptool.py
craigcomstock: I see another ticket was created for this. Ok.

build-scripts/deptool.py
craigcomstock: For the tables I guess. Makes some sense but still I wonder if we could grab it from the deps/<dep> dir/files as well. There is a ticket I think about this in general so good as-is.

larsewi: It would also be nice if we did not have to maintain this list when ever we add a new dependency.

larsewi: Also for all of these the "non-human" names are equal to the "human" names if you do `str.lower()` on it. Maybe you could keep only a list of the human names, and do `lower()` when ever you need the other?

olehermanse:
@craigcomstock we need markdown tables for release blog posts and for the README in this repo.

olehermanse (approved):

PR #1628, opened 2025-03-21
ENT-12185: Moved CFE_ROBOT user creation from MP cli task to shell script & Print setup code in the postinstall scripts

aleksandrychev:
ChangeLog: Title

Ticket: ENT-12185

together: 
 - https://github.com/cfengine/mission-portal/pull/2782
 - https://github.com/cfengine/nova/pull/2402
 - https://github.com/cfengine/buildscripts/pull/1628
 - https://github.com/cfengine/system-testing/pull/603

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -1108,6 +1108,11 @@ else
     cf_console platform_service cfengine3 start
 fi
 
+if ! is_upgrade; then
+  wait_for_cf_postgres || failure=1
+  cf_console echo "$("$PREFIX/bin/cf-hub" --new-setup-code)"
+fi
+
 rm -f "$PREFIX/UPGRADED_FROM.txt"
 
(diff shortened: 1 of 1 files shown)


craigcomstock (commented):
even better would be to move this into it's own script and call that.

Currently we have docs about how to do this but would be nice to codify it and make it "easy". Also add calling the externalized script to @olehermanse 's `cfengine` command!

craigcomstock (approved):

craigcomstock:
> even better would be to move this into it's own script and call that.
> 
> Currently we have docs about how to do this (https://docs.cfengine.com/docs/3.24/web-ui-hub_administration-reset-admin-creds.html#internal-credentials) but would be nice to codify it and make it "easy". Also add calling the externalized script to @olehermanse 's `cfengine` command!

I think this PR is a good incremental change. Approved!

craigcomstock (approved):

PR #1632, opened 2025-03-24
ENT-12711: Rename workflow for automatic dependency updates on 3.24.x

larsewi:
For some reason the workflow is only triggered on the `master` branch.
It could be a scheduled workflow is only triggered once, independent on
which branch it exists on. Hence I will try to rename the workflow, so
that it in theory becomes a different one.

Ticket: ENT-12711


olehermanse (approved):

craigcomstock (approved):
yeah, let's give it a try.

PR #1654, opened 2025-03-31
Fixed edge case when upgrading libxml2 dependency

larsewi:
- **Fixed edge case when updating libxml2**
- **Added debug log message for printing URL to be fetched**

Back-ported to https://github.com/cfengine/buildscripts/pull/1655


vpodzime (approved):
Just the same comment regarding comments as in #1655

PR #1655, opened 2025-03-31
Fixed edge case when upgrading libxml2 dependency (3.24.x)

larsewi:
- **Revert "Rename workflow for automatic dependency updates on 3.24.x"**
- **Fixed edge case when updating libxml2**
- **Added debug log message for printing URL to be fetched**

Back-ported from https://github.com/cfengine/buildscripts/pull/1654

### DIFF
--- a/.github/workflows/update-deps.py
+++ b/.github/workflows/update-deps.py
@@ -181,6 +181,12 @@ def update_version_numbers(root, pkg_name, old_version, new_version):
         os.path.join(root, DEPS_PACKAGING, pkg_name, "source"),
     ]
     for filename in filenames:
+        if filename.endswith(os.path.join("libxml2", "source")):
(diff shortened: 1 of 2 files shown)


.github/workflows/update-deps.py
```diff
@@ -181,6 +181,12 @@ def update_version_numbers(root, pkg_name, old_version, new_version):
         os.path.join(root, DEPS_PACKAGING, pkg_name, "source"),
     ]
     for filename in filenames:
+        if filename.endswith(os.path.join("libxml2", "source")):
+            """Special case for libxml2: The patch number is left out from the
+            URL of the source file"""
```
vpodzime, line 186: `#` is used for comments in Python :wink:

vpodzime (approved):
Looks good to me otherwise.

PR #1658, opened 2025-03-31
Added prefix ./ to policy file reference in build host setup script

craigcomstock:
When I tried using setup-cfengine-build-host.sh on a rhel-8 machine I guess dirname $0 didn't include a ./ so cf-agent could not find the policy file in the relative location.

Ticket: none
Changelog: none


cf-bottom:

PR #1663, opened 2025-04-08
Added 3.24.x branch to scheduled dependency updates workflow

aleksandrychev:
GitHub docs say that scheduled workflows will only run on the default branch.

This commit:
- Adds branch matrix to run updates on both master and 3.24.x branches
- Configures branch-specific settings (major bumps for master, minor for 3.24.x)

Ticket: ENT-12711


craigcomstock (approved):
yeah, let's give it a try!

craigcomstock:
If you find this works we need to add also 3.21.x to the matrix.

PR #1687, opened 2025-04-22
GH Actions: Removed contents write permission

olehermanse:
Hoping that pull-requests is enough and removing contents
limits access to other things (like releases).


craigcomstock (approved):

olehermanse:
This was needed, so reverting.

PR #1717, opened 2025-05-19
Support running hub in a non-privileged non-systemd container

craigcomstock:
- Added enterprise hub and client docker image builds
- libre
- fixup client
- wip
- libre
- libre
- works mostly except that port 8443 doesnt seem to respond when apache is running on that port
- restore post-build-setup script items for docker case
- Added better check for systemd presence


craigcomstock:
nope. don't need this PR (yet)

PR #1734, opened 2025-06-02
ENT-13025: Switched from ls to find to determine if directory is empty

nickanderson:
Ticket: ENT-13025

### DIFF
--- a/packaging/common/cfengine-hub/preinstall.sh
+++ b/packaging/common/cfengine-hub/preinstall.sh
@@ -24,13 +24,15 @@ fi
 
 test -z "$BACKUP_DIR" && BACKUP_DIR=$PREFIX/state/pg/backup
 
-if [ -d "$BACKUP_DIR" ] && [ -n "$(ls -A "$BACKUP_DIR")" ]; then
-    # If the backup directory exists and is not empty we don't want to proceed,
-    # that stale data can cause database migration problems.
-    cf_console echo "Backup directory $BACKUP_DIR exists and is not empty."
-    cf_console echo 'Please remove it, clean it, or set a different directory by setting a BACKUP_DIR env variable, like this:'
-    cf_console echo 'export BACKUP_DIR=/mnt/plenty-of-free-space'
-    exit 1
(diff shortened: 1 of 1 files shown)


larsewi (approved):

nickanderson:
Just came across this old pr thats been aproved since june 2025 I see the ticket itself was rejected. @larsewi @craigcomstock do you want this change ?

packaging/common/cfengine-hub/preinstall.sh
```diff
@@ -24,13 +24,15 @@ fi
...
-    # If the backup directory exists and is not empty we don't want to proceed,
-    # that stale data can cause database migration problems.
-    cf_console echo "Backup directory $BACKUP_DIR exists and is not empty."
-    cf_console echo 'Please remove it, clean it, or set a different directory by setting a BACKUP_DIR env variable, like this:'
-    cf_console echo 'export BACKUP_DIR=/mnt/plenty-of-free-space'
-    exit 1
+# If the backup directory exists inspect it more closely
+if [ -d "$BACKUP_DIR" ]; then
+    # If the backup directory is not empty we don't want to continue.
+    if find "$BACKUP_DIR" -maxdepth 1 -mindepth 1 -print -quit | grep -q .; then
```
craigcomstock, line 30: What is the advantage? I don't see why this is better.

larsewi (approved):
The `ls` command is known to be unreliable across different implementations. However, in this case it does not really matter, since we are not actually trying to parse the output.

PRO: this would silence a `shellcheck` warning. CON: slightly less readable.

packaging/common/cfengine-hub/preinstall.sh
craigcomstock: I agree that find is probably better but it is not more readable. For that I might suggest

```
if [ "$(du -a "$BACKUP_DIR" | wc -l)" -gt "1" ]; then
```

`du` (disk usage) is an old-timer but does make some sense and hopefully is stable. :p

PR #1742, opened 2025-06-09
Added determining base_ref on push triggers from github in build-using-buildscripts.yml workflow (3.24)

craigcomstock:
github.base_ref is not available for on:push triggers, only on:pull_request so we must find the base_ref ourselves.

Ticket: ENT-13038
Changelog: none


cf-bottom (changes requested):
Please use a company e-mail instead of c***@unreasonablefarm.org

PR #1752, opened 2025-06-13
Force all jenkins build hosts to use jdk 21

craigcomstock:
On older hosts we were getting away with jdk 11 but that isn't sufficient for impending jenkins upgrade.

This needs to hold-off while jenkins-dev is still in testing.

Ticket: ENT-13033
Changelog: none


cf-bottom:

PR #1762, opened 2025-06-20
functions: Run command and print on failure

larsewi:
Created a function to run a command and only print the output in case of
failure. This can greatly reduce noice from verbose scripts and commands
such as those related to autotools.

### DIFF
--- a/build-scripts/functions
+++ b/build-scripts/functions
@@ -596,5 +596,25 @@ rm_if_empty()
     fi
 }
 
+# Use this function on verbose commands to silence the output unless it returns
+# a non-zero exit code
+run_and_print_on_failure()
+{
+    local temp_output_file
+    temp_output_file=$(mktemp)
+    local exit_code=0
+    if "$@" > "$temp_output_file" 2>&1; then
+        : # NOOP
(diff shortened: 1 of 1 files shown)


build-scripts/functions
```diff
@@ -596,5 +596,24 @@ rm_if_empty()
     fi
 }
 
+# Use this function on verbose commands to silience the output unless it returns
```
vpodzime, line 599: *silence*

build-scripts/functions
```diff
@@ -596,5 +596,24 @@ rm_if_empty()
...
+# a non-zero exit code
+run_and_print_on_failure()
+{
+    local temp_output_file=$(mktemp)
+    local exit_code=0
+    if ($@ > "$temp_output_file" 2>&1); then
```
vpodzime, line 605: I don't think the subshell is required here. If multiple commands are to be supported, the redirects should be done from the whole subshell.

vpodzime (commented):
Looks good to me otherwise.

vpodzime (approved):
Looks good to me.

PR #1769, opened 2025-06-20
ENT-13030: fixup ubuntu 24 buildhost policy and added containers scripts to test the policy

craigcomstock:
- **Added ci scripts to use buildah to build upgraded platform images and then buildhost images**
- **Adjusted buildhost policy for ubuntu-24 regarding libncurses**
- **Adjusted several ci components to work well with containers**


ci/buildhost-images.sh
```diff
@@ -0,0 +1,41 @@
...
+  built=$platform-buildhost
+  if buildah images --format '{{.Name}}:{{.Tag}}' | grep $built; then
+    echo "container image $built already exists, remove if you want to rebuild"
+    continue
+  fi
+
+  buildah rm $c || true
+  buildah --name $c from $base
+  buildah copy $c . /buildscripts/ci
+  buildah run $c apt install -y procps wget sudo
```
nickanderson, line 23: yep all debain platforms in that list

ci/cfengine-build-host-setup.cf
```diff
@@ -148,6 +155,9 @@ bundle agent cfengine_build_host_setup
       "systemssl_build_host" expression => fileexists("/etc/cfengine-systemssl-build-host.flag");
       "bootstrap_pr_host" expression => fileexists("/etc/cfengine-bootstrap-pr-host.flag");
       "containers_host" expression => fileexists("/etc/cfengine-containers-host.flag");
+      "systemd_running" expression => returnszero("systemctl is-system-running", "useshell");
+      "not_in_container" expression => not(fileexists("/etc/cfengine-in-container.flag")),
+        comment => "We use an explicit flag file that we control to avoid ambiguity about whether we are in a container or not.";
```
nickanderson, line 160: I like the comment here that is explicit about the intent.

ci/setup-cfengine-build-host.sh
```diff
@@ -2,7 +2,7 @@
 shopt -s expand_aliases
 
 # TODO get latest LTS dynamically
-CFE_VERSION=3.24.2
+CFE_VERSION=3.26.0
```
nickanderson, line 5: TO DONE?
`cf-remote list | awk '/Using / {print $2}'` or `cf-remote --version master list | awk '/Available releases: master,/ {print $4}' | sed 's/,$//'`

nickanderson (approved):

ci/buildhost-images.sh
craigcomstock: right, would need adjustment when adding non-debians.

PR #1773, opened 2025-06-23
ENT-12600: clean-results: Documented script with comments

larsewi:
Ticket: ENT-12600


craigcomstock (commented):
where is this script even used? it does not look familiar to me.

larsewi:
> where is this script even used? it does not look familiar to me.

It's used in the build-remote script

olehermanse (approved):

PR #1802, opened 2025-07-16
Adjusted linux-install-jdk21.sh to be stable at latest-1 version

craigcomstock:
This is needed because /latest is unstable and we want to ensure stability and check checksums that we collect manually or eventually through pull request automation for human confirmation.

Ticket: ENT-13114
Changelog: none


nickanderson (approved):
LGTM

PR #1813, opened 2025-07-24
build-scripts/functions: added logging functions

larsewi:
The main motivation for these functions is for them to include the
filename in the output. Hence, when things stop working, you'll know
exactly where to look.

Example file:
```sh
#!/bin/sh
. "$(dirname "$0")/functions"
log_debug foo bar baz
log_error bogus
```

Example output:
```
$ ./test.sh
test.sh: Debug: foo bar baz
test.sh: Error: bogus
```

### DIFF
--- a/build-scripts/functions
+++ b/build-scripts/functions
@@ -617,4 +617,14 @@ run_and_print_on_failure()
     return $exit_code
 }
 
+log_debug()
+{
+    echo "$(basename "$0"): Debug:" "$@"
+}
+
+log_error()
+{
+    echo "$(basename "$0"): Error:" "$@" >&2
+}
+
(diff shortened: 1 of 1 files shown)


build-scripts/functions
```diff
@@ -617,4 +617,14 @@ run_and_print_on_failure()
     return $exit_code
 }
 
+log_debug()
+{
+    echo "$(basename "$0"): Debug:" "$@"
```
vpodzime, line 622: I'd do lowercase `debug:` and `error:` as that's more common.

vpodzime (approved):
Looks good to me either way. Lovely!

PR #1814, opened 2025-07-24
ENT-12600: Removed scripts related to building docs

larsewi:
configure-docs supposedly runs configure with --enable-docs in core.
However, there is no such option:

```
$ ./configure --help | grep docs
$ grep docs configure.ac
```

Hence, these scripts do nothing.

Ticket: ENT-12600


craigcomstock (approved):
yeah, stale!

PR #1840, opened 2025-08-13
ENT-13081: detect-environment: Fixed Ubuntu 18 thinking that it is 22 (3.21.x)

larsewi:
- **detect-environment: Fixed Ubuntu 18 thinking that it is 22**


larsewi:
It appears that the AMI was in fact Ubuntu 22

PR #1854, opened 2025-08-19
prepare testmachine chroot

larsewi:
- **prepare-testmachine-chroot: Documented script**
- **prepare-testmachine-chroot: Fixed shellcheck issues**
- **prepare-testmachine-chroot: deduped some repetitive code**

### DIFF
--- a/build-scripts/prepare-testmachine-chroot
+++ b/build-scripts/prepare-testmachine-chroot
@@ -1,6 +1,6 @@
 #!/bin/sh -x
 
-. `dirname "$0"`/functions
+. "$(dirname "$0")"/functions
 . detect-environment
 . compile-options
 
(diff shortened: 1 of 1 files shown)


build-scripts/prepare-testmachine-chroot
```diff
@@ -54,12 +54,11 @@ if [ $RETURN_CODE -ne 0 ] && [ $RETURN_CODE -ne 24 ]; then
 fi
 
 # Create and set permissions for essential directories inside chroot
-sudo mkdir -p "${CHROOT_ROOT}tmp"
+for d in tmp var/tmp run root; do
+    sudo mkdir -p $d
```
craigcomstock, line 58: s/$d/"$d"/ yeah? did you run shellcheck again?

larsewi: Yup, I think `shellcheck` does not complain because it knows that there are no spaces in `d`

larsewi:
> I wonder if we shouldn't add a make static-check target/github/jenkins task to run shellcheck on all these scripts?!

Yes, we should (see [ENT-13181](https://northerntech.atlassian.net/browse/ENT-13181)).

build-scripts/prepare-testmachine-chroot
craigcomstock: funny, you could add them just fine... indeed, it doesn't complain when the values contain no spaces. I don't like it. ;) Oh well.

craigcomstock (approved):

PR #1859, opened 2025-08-22
Fixed difference of branch name in documentation without .x suffix and other repositories with .x suffix

craigcomstock:
Also refactored how we find the upstream name from jenkins environment variables.

Ticket: ENT-12581
Changelog: none

### DIFF
--- a/ci/create-revisions-file.sh
+++ b/ci/create-revisions-file.sh
@@ -33,6 +33,22 @@ if [ -f pull-request-body ]; then
   done
 fi
 
+# figure out what our upstream branch (master, 3.24.x, etc) is from jenkins environment variables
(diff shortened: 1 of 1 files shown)


ci/create-revisions-file.sh
```diff
@@ -33,6 +33,22 @@ if [ -f pull-request-body ]; then
   done
 fi
 
+# figure out what our upstream branch is from jenkins environment variables
+# PR_BASE - from jenkins ${pullRequest.base}
```
nickanderson, line 37: For mortals like me, examples of the value are appreciated.

nickanderson (approved):
alright

PR #1866, opened 2025-08-27
Two fixes about share/gui in packaging ENT-12658, ENT-13161 (3.21)

craigcomstock:
- **Aligned Hub package perms for share/GUI on EL and Debian platforms with the MPF**
- **Marked config.php files in share/GUI folder as configuration files in packages**


craigcomstock:
@cf-bottom jenkins please

PR #1869, opened 2025-08-29
Grep for warnings in run_and_print_on_failure function

larsewi:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/12588/)

### DIFF
--- a/build-scripts/functions
+++ b/build-scripts/functions
@@ -581,10 +581,17 @@ run_and_print_on_failure() {
     # shellcheck disable=SC3043
     local exit_code=0
     if "$@" >"$temp_output_file" 2>&1; then
-        : # NOOP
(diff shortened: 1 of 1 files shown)


cf-bottom:

larsewi:
Ping @craigcomstock @olehermanse

build-scripts/functions
```diff
@@ -623,10 +623,15 @@ run_and_print_on_failure()
 
     local exit_code=0
     if "$@" > "$temp_output_file" 2>&1; then
-        : # NOOP
+        if grep -q "warning:" "$temp_output_file"; then
```
craigcomstock, line 626: seems like we would absolutely want also error and/or fail in the grep as well as maybe a few lines of context with `-C 2`

PR #1881, opened 2025-09-02
Added bullseye to list of hosts to use archive.debian.org

craigcomstock:
bullseye-backports is only at archive.debian.org so use that instead of the standard locations.

Ticket: ENT-13140
Changelog: none


cf-bottom:
Thank you for submitting a pull request! Maybe @larsewi can review this?

PR #1892, opened 2025-09-04
Adjust hub packaging scripts in cases where no files are found during upgrade (3.24)

craigcomstock:
Added --no-run-if-empty option to some find | xargs commands.

Ticket: ENT-13214
Changelog: title
(cherry picked from commit 224d4b705977ccd4c5745ce67ac8d5477ddfe185)


craigcomstock:
CI is good enough, let's go.

PR #1894, opened 2025-09-05
Fixed prepare-testmachine-chroot script quoting so that dd does not create an infinitely sized /dev/urandom in chroot cases

craigcomstock:
Ticket: ENT-13216
Changelog: none


nickanderson (approved):

craigcomstock:
no cherry picks needed as the LTS branches don't have formatting changes that broke this. 🏆

PR #1910, opened 2025-09-22
ENT-12600: unpack-tarballs: documented/refactored script

larsewi:
- **unpack-tarballs: replaced echo's with logging functions**
- **unpack-tarballs: Documented script**
- **unpack-tarballs: added a debug log message**
- **unpack-tarballs: decreased verbosity of script**


larsewi:
@cf-bottom Jenkins please :)

PR #1911, opened 2025-09-22
test: documented/refactored script

larsewi:


larsewi:
@cf-bottom Jenkins please :)

PR #1945, opened 2025-09-26
Revert "Updated Java Development Kit to 21.0.8"

craigcomstock:
This reverts commit 5d24969295d5c9c3b4976b3d0e05968d28ac7dd9.

We need latest minor -1, so 21.0.8 is latest but not in /archive/ so we need to stick with 21.0.7 for now.


olehermanse:
@craigcomstock I think that was a fluke the download works for me; https://download.oracle.com/java/21/archive/jdk-21.0.8_linux-aarch64_bin.tar.gz

PR #1962, opened 2025-10-02
ENT-12598: build-environment-check: deleted script

larsewi:
The list of wanted/unwanted packages is outdated. It only handles redhat
and debian based systems. Nowadays we use the
cfengine-build-host-setup.cf as the single source of truth. It does
exactly what we want, it promises that the build host is setup
correctly. Hence, it's time we let this script go.

Ticket: ENT-12598


larsewi:
@cf-bottom Jenkins please :)

PR #1989, opened 2025-10-17
ENT-13277: Adjusted detection of systemd in package scriptlets to handle more valid states

craigcomstock:
In the case where a package is installed but not bootstrapped several of the components: cf-execd, cf-monitord, cf-serverd will not be able to start due to lack of ${sys.inputs}/promises.cf.
This will cause the `systemctl is-system-running` command to return `degraded` and a non-zero exit code.

Previously this would cause the cfengine3 service to not be enabled as part of the package install and so on reboots services would not be running.

Ticket: ENT-13277
Changelog: title

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -29,7 +29,7 @@ systemctl restart cfengine3"
   fi
 fi
 
-if command -v systemctl 2>/dev/null && systemctl is-system-running; then
+if use_systemd; then
   # This is important in case any of the units have been replaced by the package
   # and we call them in the postinstall script.
   if ! /bin/systemctl daemon-reload; then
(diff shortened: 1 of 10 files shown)


craigcomstock:
try centos-7 only jenkins run: [Build Status](https://ci.cfengine.com/job/pr-pipeline/12846/)

craigcomstock:
beautiful green! will manually test/inspect.

nickanderson (approved):
OK, so use systemd if systemctl  exists and if systemctl is-system-running returns a non empty string that is not offline or unknown.

craigcomstock:
okee, try with my refactor with exotics but skipping acceptance tests that are flaky on exotics: [Build Status](https://ci.cfengine.com/job/pr-pipeline/12858/)

larsewi (approved):

craigcomstock:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/12871/)

craigcomstock:
I had to refactor a bit as the script-common.sh wasn't so "common" aka the (rpm|deb)-script-common.sh scripts relied on bits inside there but script-common.sh relied on bits inside of them so I created a new script: script-header.sh which comes first and contains the use_systemd function to use in many places.

craigcomstock:
this was merged to master before 3.27.0 so only picked to 3.24.x: https://github.com/cfengine/buildscripts/pull/2192

PR #2008, opened 2025-10-29
ENT-12603: autogen: removed ./ prefix when executing other scripts

larsewi:
It's not needed, and becomes problematic when executing build-remote
using an absolute path.

Ticket: ENT-12603


larsewi:
@cf-bottom Jenkins please :)

craigcomstock (approved):

PR #2009, opened 2025-10-30
ENT-12603: build-remote: adjust destination directory

larsewi:
The previous path would cause build-remote to try to make the directory `/output/`. `/` is owned by root. Hence, permissions is denied. I think it makes more sense to put these files in `$BASEDIR/output/`.

These scripts are only used by build-remote. Hence, this change should be rather safe.

Ticket: ENT-12603

### DIFF
--- a/build-scripts/clean-results
+++ b/build-scripts/clean-results
@@ -27,6 +27,6 @@
 # The `xargs` puts each line as the argument of `rm -rf` to remove the remaining
 # files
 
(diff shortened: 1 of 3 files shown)


larsewi:
@cf-bottom Jenkins please :)

olehermanse (approved):

aleksandrychev (approved):

PR #2100, opened 2026-01-28
packaging/scripts: Fix shebang

btriller:
Include script header after type specific scriptlet to ensure type specific shebang is used.

Fixes: 19ccbed5 ("Adjusted detection of systemd in package scriptlets to handle more valid states")


craigcomstock (changes requested):
I still don't see any mention of how the as-is code causes a problem. It would be good to mention what caused you to work on this change in the first place. I assume there was a problem you encountered?

PR #2182, opened 2026-04-14
feature: added build host setup handling cfengine_role env var for policy use

craigcomstock:
We can add this as a platform specific env var in jenkins config to trigger different types of build hosts.
In this case we are enabling our mingw build host setup.

Ticket: ENT-13862
Changelog: none

### DIFF
--- a/ci/setup-cfengine-build-host.sh
+++ b/ci/setup-cfengine-build-host.sh
@@ -2,6 +2,11 @@
 shopt -s expand_aliases
 thisdir="$(dirname "$0")"
 
+# handle env cfengine_role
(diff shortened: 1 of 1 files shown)


nickanderson (approved):
Just a bit curious for what those values might be cfengine-hub cfengien-client or cfengine-JiraLabel ?

craigcomstock:
> Just a bit curious for what those values might be cfengine-hub cfengien-client or cfengine-JiraLabel ?

different labels to drive different types of build host privisioning: mingw, pipeline, containers, etc.. just a one-value so far is what I need.

PR #2189, opened 2026-04-22
Refresh base image SHAs in platforms.json via weekly workflow

larsewi:
Adds a new mechanism for keeping `base_image_sha` in `platforms.json` in sync with upstream Docker Hub digests.

- `build-in-container.py` gains an `--update-sha` mode that queries Docker Hub for the current OCI index digest of each platform's `base_image` and rewrites `base_image_sha` accordingly.
- Adds `.github/workflows/update-base-image-shas.yml`, modeled on `update-base-images.yml`, which runs weekly and opens a PR with any refreshed digests. Reviewers: larsewi, craigcomstock.
- Small refactor: hoisted the repeated `Path(__file__).resolve().parent / "platforms.json"` into a `CONFIG_PATH` constant.

Rationale for keeping this separate from `update-deps.yml` (which also refreshes inputs):
- `platforms.json` only exists on `master`, not on the release branches `update-deps.yml` targets, so bolting it on there would require a branch guard.
- `base_image_sha` bumps benefit from their own reviewable PR rather than mixing with package-version bumps.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/build-in-container.py
+++ b/build-in-container.py
@@ -189,6 +189,60 @@ def update_platform_versions(platform_name=None):
     config_path.write_text(json.dumps(config, indent=2) + "\n")
 
 
+def latest_base_image_digest(base_image):
+    """Fetch current manifest digest from Docker Hub for a base image."""
+    # Docker Hub's v2 API path requires a namespace. Official images (ubuntu,
+    # debian, ...) live under "library/".
+    repo, tag = base_image.rsplit(":", 1)
+    repo = f"library/{repo}"
+
+    # The v2 API requires a bearer token even for anonymous public pulls.
+    token_url = (
+        "https://auth.docker.io/token"
+        f"?service=registry.docker.io&scope=repository:{repo}:pull"
+    )
+    token = json.loads(urllib.request.urlopen(token_url).read())["token"]
+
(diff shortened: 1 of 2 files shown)


build-in-container.py
```diff
@@ -189,6 +189,60 @@ def update_platform_versions(platform_name=None):
...
+def latest_base_image_digest(base_image):
+    """Fetch current manifest digest from Docker Hub for a base image."""
+    # Docker Hub's v2 API path requires a namespace. Official images (ubuntu,
+    # debian, ...) live under "library/".
+    repo, tag = base_image.rsplit(":", 1)
+    repo = f"library/{repo}"
+
+    # The v2 API requires a bearer token even for anonymous public pulls.
+    token_url = (
+        "https://auth.docker.io/token"
```
craigcomstock, line 201: I thought we were using ghcr.io/cfengine?

https://github.com/cfengine/buildscripts/blob/master/build-in-container.md#container-registry

craigcomstock (approved):

build-in-container.py
larsewi: Yeah, so we grab an image from docker, install our build dependencies, then we push that to ghcr.io/cfengine. Think of this as the base image of our base image.

PR #2192, opened 2026-04-22
Adjusted detection of systemd in package scriptlets to handle more valid states (3.24)

craigcomstock:
In the case where a package is installed but not bootstrapped several of the components: cf-execd, cf-monitord, cf-serverd will not be able to start due to lack of ${sys.inputs}/promises.cf.
This will cause the `systemctl is-system-running` command to return `degraded` and a non-zero exit code.

Previously this would cause the cfengine3 service to not be enabled as part of the package install and so on reboots services would not be running.

Ticket: ENT-13277
Changelog: title
(cherry picked from commit 19ccbed59d59ae763212e945e526f3d6a72fc137)

 Conflicts:
	packaging/common/cfengine-hub/postinstall.sh
	packaging/common/cfengine-non-hub/postinstall.sh

### DIFF
--- a/packaging/common/cfengine-hub/postinstall.sh
+++ b/packaging/common/cfengine-hub/postinstall.sh
@@ -29,7 +29,7 @@ systemctl restart cfengine3"
   fi
 fi
 
-if [ -x /bin/systemctl ]; then
+if use_systemd; then
   # This is important in case any of the units have been replaced by the package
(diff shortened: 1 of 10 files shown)


cf-bottom:
Thanks for submitting a pull request! Maybe @larsewi can review this?

larsewi (approved):

PR #2196, opened 2026-04-23
fix: cfengine-build-host-setup.cf needs to specify openssl version 3.1.4 so that suse-15 can load libopenssl-devel which is stuck at the older version 3.1.4 instead of current 3.2.3

craigcomstock:
Ticket: ENT-13999
Changelog: none


cf-bottom:
Thank you for submitting a PR! Maybe @larsewi can review this?

larsewi (approved):

craigcomstock:
This actually did not work. I reverted to using sles 15.6 and that solves the issue in a different way.

PR #2219, opened 2026-04-30
Formatted script ci/setup-cfengine-build-host.sh

larsewi:


craigcomstock (approved):
I respectfully disagree... 4 spaces is twice as much to type :p but yeah, I can use a formatter :'(

larsewi:
We can use 2 spaces @craigcomstock. But we should stick to what ever we chose. The main motivation for formatting this file was that there were a mix of spaces and tabs.

PR #2225, opened 2026-05-01
fix: ci/initialize-build-host.sh: fail if workspace cannot be cleaned up (3.27)

craigcomstock:
- **fix: add ci/fix-buildhost.sh to source /etc/profile especially for exotics which may have basic tools like ssh in odd places added to PATH in /etc/profile**
- **fix: tidy up ci/initialize-build-host.sh to be more quiet and fail if workspace cleanup fails**


craigcomstock:
trying a small build: https://ci.cfengine.com/job/pr-pipeline/13696/

PR #2233, opened 2026-05-04
GH Actions: Added condition to prevent running in forks

olehermanse:


nickanderson (approved):

olehermanse:
does not exist in other branches, no backport.

PR #2249, opened 2026-05-11
ENT-13786: Reproducible container builds

larsewi:
Make the container build path produce bit-identical packages across two cache-cleared runs.

Four commits:

1. **Export `SOURCE_DATE_EPOCH`** at the top of `build-in-container-inner.sh`. Honored by OpenSSL, Apache httpd, PHP, OpenLDAP, `dpkg-buildpackage`, and `rpmbuild`. Pins embedded build timestamps.
2. **`composer install --prefer-dist`** to avoid leaving `.git/` directories in the vendor tree. Hint, not enforced — see commit 4.
3. **PHP source patch** so `ext/phar` honors `SOURCE_DATE_EPOCH`. Upstream PHP's phar packer unconditionally embeds `time(NULL)` into manifest entries. Patches four call sites in `ext/phar/{phar,util,tar,zip}.c`. Worth submitting upstream to php-src separately so we don't carry it indefinitely.
4. **Strip composer vendor `.git/` after install.** `--prefer-dist` falls back to source when GitHub's anonymous zipball API rate-limits (60 req/hour; mission-portal has dozens of packages). Matches what CI does accidentally via `--exclude=".git"` in `build-scripts/transfer-to-buildmachine`.

```
$ sha256sum *.deb
33247d26a4b05dadfe2f20e1810c69d2e8e148e9984d4c905ae5d562496f0fd4  post1-fix.deb
33247d26a4b05dadfe2f20e1810c69d2e8e148e9984d4c905ae5d562496f0fd4  post2-fix.deb
```

### DIFF
--- a/build-in-container-inner.sh
+++ b/build-in-container-inner.sh
@@ -81,6 +81,11 @@ install_mission_portal_deps() (
         echo "Installing LDAP API PHP dependencies..."
         (cd "$BASEDIR/mission-portal/ldap" && php /usr/bin/composer.phar install --no-dev --ignore-platform-reqs --prefer-dist)
     fi
+
+    # Composer falls back to git clone when GitHub's anonymous zipball
+    # rate limit is hit, leaving non-reproducible .git directories in the
+    # vendor tree. Strip them.
+    find "$BASEDIR/mission-portal" "$BASEDIR/nova/api/http" -type d -name .git -path '*/vendor/*' -exec rm -rf {} +
 )
 
 # === Step runner with failure reporting ===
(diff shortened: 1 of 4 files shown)


craigcomstock (approved):
Is it the case that we could run the build-in-container-inner.sh script by itself on a host and build? We are putting a lot of important logic in there and I wonder if we shouldn't refactor things to make it more obvious that we could run this script "anywhere" so to speak?

larsewi:
> Is it the case that we could run the build-in-container-inner.sh script by itself on a host and build? We are putting a lot of important logic in there and I wonder if we shouldn't refactor things to make it more obvious that we could run this script "anywhere" so to speak?

Yepp, definitely deserves refactoring at some point ;)

PR #2262, opened 2026-05-20
ENT-14108: drain cf-agent in hub preremove.sh before stopping cfengine3 umbrella

larsewi:
Race: prerm stops cf-execd and cf-postgres, but a still-running cf-agent then starts cf-php-fpm — which pulls cf-postgres back in as a dependency (`Wants=cf-postgres.service`). The start fails (data dir being torn down) and `Restart=always` keeps it looping into the next install's postinst, eventually clobbering postinst's postgres so the final `pg_ctl stop` fails and dpkg aborts.

This PR drains in-flight cf-agent in prerm before the umbrella stop, so policy can't re-trigger services during teardown. See the commit message for the full mechanism.

### DIFF
--- a/packaging/common/cfengine-hub/preremove.sh
+++ b/packaging/common/cfengine-hub/preremove.sh
@@ -1,3 +1,16 @@
+# Drain in-flight cf-agent before stopping the cfengine3 umbrella: otherwise
+# a running cf-agent can re-trigger cf-php-fpm (which Wants=cf-postgres),
+# leaving cf-postgres in a Restart=always loop while the package is gone.
+if use_systemd; then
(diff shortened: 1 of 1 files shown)


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/13821/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/13821/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-13821/

larsewi:
Will move the kill to `core/misc/systemd/cf-execd.service`

PR #2287, opened 2026-06-03
ENT-14152: Reformatted with CFEngine CLI and added lint and format checks (3.27)

olehermanse:


larsewi (approved):
We don't really use `ci/cfengine-build-host-setup.cf` from other branches than master. Could consider just removing it

craigcomstock (approved):

craigcomstock:
> We don't really use `ci/cfengine-build-host-setup.cf` from other branches than master. Could consider just removing it

But in the near future we will.

PR #2338, opened 2026-07-07
ENT-13867: Added RHEL 8,9,10 support to build in container

aleksandrychev:
Ticket: ENT-13867

### DIFF
--- a/.github/workflows/build-base-images.yml
+++ b/.github/workflows/build-base-images.yml
@@ -20,6 +20,9 @@ jobs:
           - debian-11
           - debian-12
           - debian-13
+          - rhel-8
+          - rhel-9
+          - rhel-10
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
--- a/container/Dockerfile.rhel
+++ b/container/Dockerfile.rhel
@@ -1,0 +1,64 @@
+ARG BASE_IMAGE=rockylinux/rockylinux:9
+FROM ${BASE_IMAGE}
+
+# allow older packages when the newest has unmet deps (as build hosts do).
+RUN echo "best=False" >> /etc/dnf/dnf.conf
+
+# Enable CRB (crb on 9+, powertools on 8) + EPEL for the -devel/build packages.
+ARG CRB_REPO=crb
+RUN dnf install -y dnf-plugins-core epel-release \
+    && dnf config-manager --set-enabled ${CRB_REPO} \
+    && dnf clean all
+
+# Deps like zlib build without full debug symbols; disable debuginfo extraction
+# so rpm doesn't abort with "No debugging symbols". We don't ship debuginfo here.
+RUN echo '%debug_package %{nil}' > /etc/rpm/macros.cfengine-nodebug
+
+# Build toolchain (see redhat section of ci/cfengine-build-host-setup.cf).
+# "Development Tools" provides gcc, make, autotools, rpm-build, core perl, etc.
+RUN dnf groupinstall -y "Development Tools" \
+    && dnf install -y \
+        gcc-c++ pam-devel ncurses ncurses-devel expat expat-devel gettext \
+        rpm-build-libs selinux-policy selinux-policy-devel \
+        pkgconf pkgconf-pkg-config \
+        perl perl-Module-Load-Conditional perl-ExtUtils-MakeMaker perl-IPC-Cmd \
+        perl-IO-Compress \
+        python3 python3-pip python3-devel \
+        git rsync wget which psmisc unzip diffutils file sudo \
+    && dnf clean all
+
+# Hub tools: Node 20 (system nodejs too old for the node: protocol) + less.
+RUN curl -fsSL https://rpm.nodesource.com/setup_20.x | bash - \
+    && dnf install -y nodejs \
+    && dnf clean all
+RUN npm install -g less
+
+ARG PHP_MODULE_STREAM=""
+RUN if [ -n "${PHP_MODULE_STREAM}" ]; then \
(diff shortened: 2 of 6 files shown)


container/Dockerfile.rhel
```diff
@@ -0,0 +1,64 @@
+ARG BASE_IMAGE=rockylinux/rockylinux:9
+FROM ${BASE_IMAGE}
```
craigcomstock, line 2: I don't think we should duplicate in a different way what we do in the build host setup. I would be fine with choosing either policy or scripts but not both.

container/Dockerfile.rhel
```diff
@@ -0,0 +1,64 @@
+ARG BASE_IMAGE=rockylinux/rockylinux:9
```
craigcomstock, line 1: ideally we will use these to build our packages so it seems we should use ubi/rhel images and not rocky.

craigcomstock (changes requested):

container/Dockerfile.rhel
aleksandrychev: @craigcomstock I took https://github.com/cfengine/buildscripts/blob/master/container/Dockerfile.debian as reference, you mean install cfengine to install dependencies? If yes I'd like to have scripts rather than policies in this case to avoid chicken\egg problem when new platforms added.

container/Dockerfile.rhel
aleksandrychev: I looked into UBI, but its repos only contain a free redistributable subset of RHEL, and some packages we need to build (bison, flex, pam-devel, selinux-policy-devel) aren't included, see the public repo listing, e.g. no bison in the appstream: https://cdn-ubi.redhat.com/content/public/ubi/dist/ubi9/9/x86_64/appstream/os/Packages/b/ . Getting them on UBI requires RHEL entitlement certs inside the image build.

.github/workflows/build-base-images.yml
```diff
@@ -20,6 +20,9 @@ jobs:
           - debian-11
           - debian-12
           - debian-13
+          - rhel-8
```
larsewi, line 23: We might need Red Hat 7 as well

larsewi (approved):

container/Dockerfile.rhel
craigcomstock: Using policy hasn't bought us so much and caused some trouble.

I would suggest we begin refactoring to stop using that policy and instead use scripts.

The goal was to make it "easier" to manage and this has not been the result.

Feel free to make a follow-up ticket for migrating away from policy to scripts for build host setup.

container/Dockerfile.rhel
craigcomstock: That is correct. @olehermanse @nickanderson @vpodzime how would you suggest we proceed? Is the difference between rockylinux and rhel so little that we can rely on rockylinux to build?

nickanderson: So many redhat issues. Most of our users are rhel proper.  for basic building irocky might be ok  but we need regular tests against rhel for real i think

container/Dockerfile.rhel
craigcomstock: > @craigcomstock I took https://github.com/cfengine/buildscripts/blob/master/container/Dockerfile.debian as reference, you mean install cfengine to install dependencies? If yes I'd like to have scripts rather than policies in this case to avoid chicken\egg problem when new platforms added.

Also, there is a provision to build from source enough to bootstrap a new platform. If that doesn't work then we should fix it or remove it. It did work recently for rhel-10-arm fwiw. https://github.com/cfengine/buildscripts/blob/master/ci/setup-cfengine-build-host.sh#L217

container/Dockerfile.rhel
olehermanse: @craigcomstock short answer; I don't know. Long answer; we're splitting up building and testing, in theory, as long as we test on RHEL proper, I think it should be okay to build on something "similar enough".

vpodzime: > Is the difference between rockylinux and rhel so little that we can rely on rockylinux to build?

Why not CentOS Stream?

aleksandrychev: > > Is the difference between rockylinux and rhel so little that we can rely on rockylinux to build?
> 
> Why not CentOS Stream?

CentOS Stream sits ahead of RHEL. It's a rolling preview of the next RHEL minor release, so it can include newer versions of dependencies that haven't been released in RHEL yet. Rocky is a downstream (100% bug-for-bug compatible with Red Hat Enterprise Linux, as they claim).

vpodzime: Right, so RHEL sits between Rocky and CentOS Stream. I think both of those options are equally good or bad.

larsewi: Older is better than newer to avoid stuff like incompatible `glibc`

vpodzime: That never happens on RHEL.

craigcomstock (approved):
please make a follow-up QA ticket to choose policy or script for build host provisioning and refactor so we aren't copy/pasting bits around everywhere as we are now. e.g. Dockerfile.rhel should just call a script that is also called by setup-cfengine-build-host, build host packer image creation, etc...

Other than that, this PR looks good enough.

aleksandrychev:
> please make a follow-up QA ticket to choose policy or script for build host provisioning and refactor so we aren't copy/pasting bits around everywhere as we are now. e.g. Dockerfile.rhel should just call a script that is also called by setup-cfengine-build-host, build host packer image creation, etc...
> 
> Other than that, this PR looks good enough.

https://northerntech.atlassian.net/browse/ENT-14301

PR #2353, opened 2026-07-17
Moved last two needed scripts from mender-qa to buildscripts

craigcomstock:
- **Moved create_swap_file.sh and initialize-user-data.sh from mender-qa/scripts to buildscripts/ci**
- **Remove SSH port forwarding from initialization script**
- **Tidied up initialize-user-data.sh and initialize-build-host.sh scripts for move from mender-qa to buildscripts**

### DIFF
--- a/ci/cfengine-build-host-setup.cf
+++ b/ci/cfengine-build-host-setup.cf
@@ -269,6 +269,7 @@ bundle agent cfengine_build_host_setup
         expression => not(fileexists("/etc/cfengine-in-container.flag")),
         comment => "We use an explicit flag file that we control to avoid ambiguity about whether we are in a container or not.";
 
+@if minimum_version(3.23)
     # Rust is build dependency for leech2 (gate on ubuntu>=20, debian>=12, redhat>=7)
     ubuntu::
       "leech2_build_toolchain_host"
@@ -281,6 +282,7 @@ bundle agent cfengine_build_host_setup
     (redhat|centos)::
       "leech2_build_toolchain_host"
         expression => version_compare("$(sys.os_version_major)", ">=", "7");
+@endif
 
     any::
       "have_rust" expression => fileexists("/opt/rust/bin/rustc");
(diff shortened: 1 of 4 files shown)


ci/initialize-user-data.sh
```diff
@@ -0,0 +1,83 @@
...
+# Enable sudo access for jenkins.
+echo "jenkins ALL=(ALL:ALL) NOPASSWD:ALL" >> /etc/sudoers
+
+# Disable TTY requirement.
+sed -i -e 's/^\( *Defaults *requiretty *\)$/# \1/' /etc/sudoers
+
+# Copy the buildscripts repository to jenkins user.
+cp -r /root/buildscripts /home/jenkins
+# was copying build-artifacts-cache known host entry
+#cp /root/mender-qa/data/known_hosts                    /home/jenkins/.ssh/known_hosts
```
craigcomstock, line 29: This should be a TODO, check where and how this file might be managed other than here.

I did not include the data/known_hosts file from mender-qa as I think the data we want arrives in a different way already.

craigcomstock: I will check on this soon when I have newly baked CI images to test a build with.

craigcomstock: will skip this for now and adjust in a follow-up PR.

ci/cfengine-build-host-setup.cf
```diff
@@ -269,6 +269,7 @@ bundle agent cfengine_build_host_setup
         expression => not(fileexists("/etc/cfengine-in-container.flag")),
         comment => "We use an explicit flag file that we control to avoid ambiguity about whether we are in a container or not.";
 
+@if minimum_version(3.23)
```
nickanderson, line 272: Heh, I see why but wasnt expecting a macro for that.

nickanderson (approved):

ci/cfengine-build-host-setup.cf
craigcomstock: yup :)

PR #2370, opened 2026-07-24
Pin libopenssl-3-devel to installed openssl version on SUSE 15 

aleksandrychev:
The legacy libopenssl-devel metapackage hard-pins openssl to an exact version. After the golden AMI rebake updated the runtime openssl from SP6 (3.1.4) to SP7 (3.2.3), the SP6-only libopenssl-devel could no longer install without downgrading the whole openssl stack, which zypper refuses non-interactively. libopenssl-3-devel depends on libopenssl3 rather than an exact openssl version, so it matches the installed 3.2.3.

Ticket: ENT-14345

### DIFF
--- a/ci/cfengine-build-host-setup.cf
+++ b/ci/cfengine-build-host-setup.cf
@@ -123,9 +123,9 @@ bundle agent cfengine_build_host_setup
       "platform-python-devel" -> { "cfbs shebang", "ENT-11338" }
         comment => "py3_shebang_fix macro needs /usr/bin/pathfix.py from platform-python-devel package";
 
-    suse_15::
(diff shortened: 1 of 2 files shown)


aleksandrychev:
Closing in favor of fixing the root cause in infra. The SP6/SP7 openssl skew came from provision.sh adding openSUSE Leap 15.6 repos (priority 90/70) to the SLES 15 SP7 golden AMI, outranking the SP7 SLE modules (priority 99). Infra now guards that block with `[ "$ID" != "sles" ]` so SLES resolves openssl-devel from its registered SP7 modules and no downgrade is attempted. The version-pin workaround here is no longer needed.

PR #2376, opened 2026-07-27
Added MIT license

larsewi:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Northern.tech
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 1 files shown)


nickanderson (approved):
I can see this matches core.

vpodzime (commented):
The commit message is misleading -- this adds the dual-license file, not GPLv3. I'd personally choose a different license here, but that's a separate discussion.

larsewi:
> I'd personally choose a different license here, but that's a separate discussion.

@vpodzime which one do you prefer ?

vpodzime:
> @vpodzime which one do you prefer ?

AFAICT, the effects of GPL are basically void here because nobody is re-distributing this with their own changes. I'd go with the least restrictive option possible so MIT (or Apache 2.0 to align with some other repos we have).

LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2026 Northern.tech
```
vpodzime, line 3: We have no copyright for this text. (I know we make this mistake in other repos)

vpodzime (commented):
Should we also add a CLA? We shouldn't claim copyright for contributed code/changes otherwise.

LICENSE
larsewi: When you create a new repository in GitHub and choose the MIT license, this will appear here by itself. Maybe GitHub also got this wrong?

vpodzime: What? How would they know whom to add there as the copyright owner/claimer? Isn't it some template of ours? Anyway, not a big deal, like I wrote above, we have this everywhere and we should fix it (if really necessary) everywhere at once.

olehermanse: @vpodzime This is not a mistake, nor a template we made - Please take a look at the MIT license template (from OSI);

https://opensource.org/license/mit

The LICENSE file, including the Copyright line, is about the copyright (and license) of the repo / folder it is in.

They also have similar instructions for other licenses, such as the Apache License, ISC, BSD, etc.

vpodzime: > The LICENSE file, including the Copyright line, is about the copyright (and license) of the repo / folder it is in.

That's an interesting implicit extension of the usual meaning/effect, but I see. Thanks for the extra info!

larsewi:
> Should we also add a CLA? We shouldn't claim copyright for contributed code/changes otherwise.

I will look into this

larsewi:
> Should we also add a CLA?

@vpodzime I think you are the correct person to answer this :beers: We use a CLA in core which is GPL. In other repos with permissive licenses we don't use it.

vpodzime:
> We use a CLA in core which is GPL. In other repos with permissive licenses we don't use it.

It's because of the dual license, not GPL exactly, AFAICT. I guess it's not needed here because we won't need to change the license here or use two possible licenses.

PR #2395, opened 2026-08-04
ENT-14361, ENT-14362: Preparing build scripts for running build-in-container in Jenkins

larsewi:


cf-bottom:

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Sure, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14412/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14412/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14412/

build-scripts/package
```diff
@@ -9,6 +9,27 @@
...
+    _tarball="$1"
+    _dir="$2"
+
+    if [ -n "$SOURCE_DATE_EPOCH" ]; then
+        find "$_dir" -exec touch -h -d "@$SOURCE_DATE_EPOCH" {} +
+    fi
+
+    find "$_dir" | LC_ALL=C sort >"$_tarball.filelist"
+    tar --no-recursion --numeric-owner --owner=0 --group=0 \
+        -cf - -T "$_tarball.filelist" | gzip --no-name >"$_tarball"
```
aleksandrychev: `tar | gzip` throws away tar's exit status — POSIX `sh` has no `pipefail`, so `set -e` only ever sees gzip's 0. The old `tar czvf` aborted the build when tar failed; now a truncated tarball gets a sha256 written for it and ships as good.

Writing the tar to a temp file and gzipping it as a second command would keep the failure fatal.

build-in-container-inner.sh
```diff
@@ -33,23 +33,47 @@ for repo in $repos; do
...
     fi
 done
 
+# The dependency cache is reached over sftp, so the key has to be in place
+# before install-dependencies runs. It arrives on a read-only mount owned by the
+# host user, and ssh refuses a key owned by anyone but us, hence the copy.
+if [ -f /run/secrets/sftp-cache-key ]; then
+    echo "Installing dependency cache key..."
+    install -d -m 700 "$HOME/.ssh"
+    install -m 600 /run/secrets/sftp-cache-key "$HOME/.ssh/id_rsa"
```
aleksandrychev: This runs as the unprivileged `builder`, but the source is a 0600 file on a read-only mount owned by the host user — so if the host uid differs from `builder`'s, `install` can't read it and the build dies here (or at the sftp probe below). Every other mount in this script gets a `sudo chown` for exactly this reason; this one doesn't.

`sudo install -m 600 -o "$(id -u)" -g "$(id -g)" ...` would cover it.

build-in-container.py
```diff
@@ -344,11 +359,42 @@ def update_base_image_shas(platform_name=None):
...
+    # platforms into one output directory does not mix them together. The
+    # tarballs belong to no platform, so they sit beside those directories.
+    subdir = "tarballs" if args.tarballs else label
+    output_dir = Path(args.output_dir).resolve() / subdir
     cache_dir = Path(args.cache_dir).resolve()
 
+    # Start from an empty directory so that packages left by an earlier build,
+    # of another project or build type, cannot be taken for this build's output.
+    if output_dir.exists():
+        shutil.rmtree(output_dir)
```
aleksandrychev, line 401: `label` encodes platform and role, but not project or build type — so a nova build and a community build of the same platform resolve to the same subdir, and the `rmtree` means whichever runs second deletes the first one's packages.

The comment names "another project or build type" as the case being guarded against, but those are the ones that need to stay side by side. Folding `args.project`/`args.build_type` into the subdir would fix both.

build-in-container.py
```diff
@@ -589,15 +669,22 @@ def main():
             f"Building {args.project} {args.role} for {args.platform} ({args.build_type})..."
         )
 
+    # No dependencies are built for the tarballs, so they need no cache label.
+    label = (
+        None
+        if args.tarballs
+        else cache_label(args.platform, args.role, args.arch or host_docker_arch())
```
aleksandrychev: `host_docker_arch()` is the *host's* arch, not the one being built. With no `--arch` given, an amd64-only platform built on an arm64 host gets `PACKAGES_arm_64_...` — wrong buildcache namespace, and wrong output subdir. The arch the container actually runs as is what this wants.

aleksandrychev (commented):
A few blockers from reading through the diff.

build-in-container.py
larsewi: That's true, but the two never share an output directory in practice: community and enterprise are built by separate jobs, each with its own workspace. I'd like to keep the structure as close to testing-pr's as possible, so that replacing its packages with these breaks as little as possible downstream.

I will drop the "another project or build type" of the comment, since it is not precise.

larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14455/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14455/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14455/

build-in-container.md
```diff
@@ -46,6 +46,7 @@ None of the above arguments are required for `--update`.
 | ------------------ | -------------------------------- | ---------------------------------------------------------------------------------- |
 | `--output-dir`     | `./output`                       | Where to write output packages                                                     |
 | `--cache-dir`      | `~/.cache/cfengine/buildscripts` | Dependency cache directory                                                         |
+| `--sftp-key`       |                                  | Private key for the remote dependency cache. Uses local cache if not specified     |
```
craigcomstock, line 49: So in order to use this it must be a file yes? Would be nice to have it flow from mystiko as content for a local developer maybe so it is never written to a file on the build host.

build-in-container.py
```diff
@@ -425,11 +431,19 @@ def run_container(args, image_tag, source_dir, script_dir):
...
             f"JOB_BASE_NAME=label={label}",
-            "-e",
-            "CACHE_IS_ONLY_LOCAL=yes",
         ]
     )
 
+    # The remote dependency cache is reachable by publickey only, and pkg-cache
+    # aborts the build if an upload fails, so it stays off unless a key was
+    # passed. Note that the key is readable by everything the build runs,
+    # including each dependency's own build system.
```
craigcomstock, line 440: Seems we should log a follow up security ticket to isolate the key to only our trusted scripts.

build-in-container.md
```diff
@@ -23,93 +23,45 @@ specified, defaults will:
   (`~/.cache/cfengine/buildscripts`).
 - Use the current working directory for output packages (`./output/`).
 
-## Usage
+### Usage
 
+See:
```
craigcomstock, line 28: Simplifying the docs is such a good move! Nice one!

build-in-container.md
```diff
@@ -133,6 +133,17 @@ The inner script runs these steps in order:
 5. **compile** -- compiles and installs to the dist tree
 6. **package** -- creates `.deb` or `.rpm` packages
 
+## Source tarballs
+
+The core and masterfiles tarballs are built in a dedicated container with:
```
craigcomstock, line 138: Is there any provision for trying to build a platform triggering the tar balls if it has not yet been run? Aka an "ease of use" issue, maybe solved by just an error message early in a platform build.

build-scripts/install-dependencies
```diff
@@ -173,8 +173,16 @@ for dep in $DEPS; do
...
-    # Note that this revision file is created by build-scripts/autogen
-    revision=$(cat "$BASEDIR"/buildscripts/deps-packaging/revision)
+    #
+    # Note that this revision file is created by build-scripts/autogen. It holds
+    # the revision of this dependency and the time of that commit. The revision
+    # ends up in the package version and is used in the key on build cache. The
+    # time pins the timestamps embedded in the packages making them
+    # reproducible.
+    revision_file="$BASEDIR/buildscripts/deps-packaging/$dep/revision"
+    [ -f "$revision_file" ] || fatal "no revision file: $revision_file"
```
craigcomstock, line 183: I think it is important to give clear information in user facing output how the revision file should be created in case a user is working locally step by step.

container/Dockerfile.rhel7
```diff
@@ -0,0 +1,64 @@
...
+# repositories, which suits us: the package set cannot shift under a rebuild.
+RUN sed -i 's/mirror.centos.org/vault.centos.org/; /^mirrorlist/d; s/^#baseurl/baseurl/' \
+        /etc/yum.repos.d/CentOS-Base.repo \
+    && yum -y makecache
+
+# Deps like zlib build without full debug symbols; disable debuginfo extraction
+# so rpm doesn't abort with "No debugging symbols". We don't ship debuginfo here.
+RUN echo '%debug_package %{nil}' > /etc/rpm/macros.cfengine-nodebug
+
+# Build toolchain (see the redhat section of ci/cfengine-build-host-setup.cf).
```
craigcomstock, line 22: Seems like much of this should call a script shared with build host setup or be more accessible if someone wants to setup a host directly for development. Granted they can just start the py script as she'll but still feels better to keep some flexibility and separate what from how aka this is implementation locked into containers.

craigcomstock (approved):
Totally good as is. A few comments suggestions is all.

build-in-container.md
larsewi: Can you elaborate on this?

build-in-container.py
larsewi: Any ideas how we could do this?

build-in-container.md
larsewi: So what you are thinking is that the packages are dependent on the source tarball? This is not the case for build in container. But maybe it should be? This way we test that the tarballs also work on all the platforms.

build-scripts/install-dependencies
larsewi: https://northerntech.atlassian.net/browse/CFE-4710

container/Dockerfile.rhel7
larsewi: https://northerntech.atlassian.net/browse/CFE-4711

PR #2400, opened 2026-08-05
Added github hostkeys to known_hosts in setup-ci-host.sh

craigcomstock:


craigcomstock:
see https://ci.cfengine.com/job/docs-pr/228/console for how this works, more quiet

nickanderson (approved):
It looked fine to me, but claude says: this only fixes container hosts. Regular build hosts take a different path through the code and still won't have GitHub's keys.

craigcomstock:
> It looked fine to me, but claude says: this only fixes container hosts. Regular build hosts take a different path through the code and still won't have GitHub's keys.

Claude is short-sighted and doesn't understand the scripts fully. The setup-cfengine-build-host.sh script only calls setup-ci-host.sh in case of containers. The rest are managed by the policy which includes provisioning the hostkeys to known_hosts. :)

PR #2401, opened 2026-08-05
Revert use of -T /tmp in initialize-build-host due to disk full troubles

craigcomstock:
On some hosts, /tmp is not large enough to hold the transfer.
This was introduced in commit 8b51092b10f5d178e2f4a1bd4312ec163eb4588e with the comment: Try to avoid temporary rsync files from other hosts while copying.

Ticket: ENT-14400
Changelog: none


cf-bottom:

nickanderson (approved):

PR #2402, opened 2026-08-06
setup-ci-host.sh: Replaced packages alias with function

larsewi:
Aliases are expanded when a function body is parsed, so the packages alias defined later during OS discovery was never resolved inside install-packages.

### DIFF
--- a/ci/setup-ci-host.sh
+++ b/ci/setup-ci-host.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 set -e
-shopt -s expand_aliases
 thisdir="$(dirname "$0")"
 
 _packages="" # a space separated list of packages to install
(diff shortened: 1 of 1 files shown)


larsewi:
@cf-bottom Jenkins please :)

cf-bottom:
Alright, I triggered a build:

[Build Status](https://ci.cfengine.com//job/pr-pipeline/14400/)

**Jenkins:** https://ci.cfengine.com/job/pr-pipeline/14400/

**Packages:** http://buildcache.cfengine.com/packages/testing-pr/jenkins-pr-pipeline-14400/

larsewi:
> Aliases are expanded when a function definition is read, not when the function is executed, because a function definition is itself a command. 

- Ref. https://www.gnu.org/software/bash/manual/html_node/Aliases.html

craigcomstock (approved):

PR #2413, opened 2026-08-10
Build on opensuse without pam due to conflicting libcrypt versions vendored and in distribution

craigcomstock:
Ticket: ENT-14406
Changelog: title

### DIFF
--- a/build-scripts/configure
+++ b/build-scripts/configure
@@ -111,6 +111,10 @@ case "$BUILD_TYPE" in
     ;;
 esac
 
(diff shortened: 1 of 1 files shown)


craigcomstock:
[Build Status](https://ci.cfengine.com/job/pr-pipeline/14447/)

craigcomstock:
We don't need this. It is a simpler fix to include one cipher that suse pam modules rely on. See https://github.com/cfengine/buildscripts/pull/2417

PR #2415, opened 2026-08-11
Added subordinate id ranges for jenkins on CI container hosts

nickanderson:
Adds subordinate id ranges for the jenkins user on CI container hosts, so rootless buildah and podman stop dropping setuid bits while extracting image layers. Existing storage is remapped and affected images dropped so they are rebuilt.

To fix the docs build, need to re-run setup ci host on the physical hosts.

### DIFF
--- a/ci/setup-ci-host.sh
+++ b/ci/setup-ci-host.sh
@@ -25,6 +25,28 @@ function file-line()
   fi
 }
 
+# Replaces an existing range for the user rather than appending a second one.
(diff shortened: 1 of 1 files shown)


craigcomstock (approved):
yeah, makes sense.

nickanderson:
Related to this pr build failure: https://github.com/cfengine/documentation/pull/3731#issuecomment-5226905876

PR #2425, opened 2026-08-12
ENT-14361: Added --label to build by Jenkins build label

larsewi:
The `build-in-container.py` script already knows how a Jenkins label maps to a platform. Hence, an option to build based on a label makes it so that a Jenkins job does not also have to know this mapping.

With this change, adding a platform will only require editing `platforms.json` and `labels.txt`.

### DIFF
--- a/build-in-container.md
+++ b/build-in-container.md
@@ -11,6 +11,9 @@ only Docker and Python 3 on the host.
 
 # Build a nova hub release package for Debian 12
 ./build-in-container.py --platform debian-12 --project nova --role hub --build-type RELEASE
+
+# Alternatively, use Jenkins labels (see build-scripts/labels.txt)
+./build-in-container.py --label PACKAGES_HUB_x86_64_linux_debian_12 --project nova --build-type RELEASE
 ```
 
 In the examples above, we run the script from inside `buildscripts/` (with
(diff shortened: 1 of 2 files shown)


build-in-container.md
```diff
@@ -11,6 +11,9 @@ only Docker and Python 3 on the host.
 
 # Build a nova hub release package for Debian 12
 ./build-in-container.py --platform debian-12 --project nova --role hub --build-type RELEASE
+
+# Alternatively, use Jenkins labels (see build-scripts/labels.txt)
+./build-in-container.py --label PACKAGES_HUB_x86_64_linux_debian_12 --project nova --build-type RELEASE
```
craigcomstock, line 16: For this case of _HUB_ you don't need project nova, that is required.

I guess strangely, we don't have a distinction between community and enterprise platform labels. ok.

Might be interesting to consider changing this to `--edition (community|enterprise)`

craigcomstock (approved):

craigcomstock (approved):

build-in-container.md
larsewi: - https://northerntech.atlassian.net/browse/CFE-4712
- https://northerntech.atlassian.net/browse/CFE-4713

PR #2434, opened 2026-08-14
ENT-14361: Added a Jenkins pipeline that builds packages in containers

larsewi:
The CFEngine organization folder scans the org for a Jenkinsfile in the repository root, so this file alone makes a job of every branch and every pull request. Nothing in infra defines the job.

Ticket: ENT-14361

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,252 @@
+// build-in-container -- CFEngine packages built in containers on the
+// CONTAINER_PACKAGES_* docker hosts.
+//
+// Jenkins finds this file itself: the CFEngine organization folder scans the
+// org for a Jenkinsfile in the repository root and makes a job of every branch
+// and every pull request, each running its own copy of this file. Nothing in
+// infra defines the job.
+
+// Returns the labels to build. build-scripts/labels.txt decides what a branch
+// builds. readTrusted reads it from the revision this job is for, without
+// cloning the repo.
+def selectedLabels() {
+  def sel = readTrusted('build-scripts/labels.txt').readLines()
+      .collect { it.trim() }
+      .findAll { it.startsWith('PACKAGES_') && it ==~ params.LABEL_FILTER }
+  if (!sel) {
+    error "LABEL_FILTER '${params.LABEL_FILTER}' matches no label in build-scripts/labels.txt."
+  }
+  return sel
+}
+
+// Returns the architecture of the node a label has to build on.
+def archOf(String label) {
+  return label.contains('_arm_64') ? 'arm64' : 'amd64'
+}
+
+// Return the repos we need to check out.
+def reposFor(String project) {
+  def repos = ['buildscripts', 'core', 'masterfiles']
+  if (project == 'nova') { repos.addAll(['enterprise', 'nova', 'mission-portal']) }
+  return repos
+}
+
+// The revision to use when a repo's parameter is empty. Normally the branch
+// this job is for. In a pull request job buildscripts gets the pull request.
+// The other repos get the branch it targets.
+//
+// BASE_BRANCH wins over the job's branch. pr-pipeline passes it, and its matrix
+// jobs build it, so this job has to agree with them.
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,252 @@
...
+
+// Returns the labels to build. build-scripts/labels.txt decides what a branch
+// builds. readTrusted reads it from the revision this job is for, without
+// cloning the repo.
+def selectedLabels() {
+  def sel = readTrusted('build-scripts/labels.txt').readLines()
+      .collect { it.trim() }
+      .findAll { it.startsWith('PACKAGES_') && it ==~ params.LABEL_FILTER }
+  if (!sel) {
+    error "LABEL_FILTER '${params.LABEL_FILTER}' matches no label in build-scripts/labels.txt."
```
craigcomstock, line 17: oh I love this so much!

Jenkinsfile
```diff
@@ -0,0 +1,252 @@
...
+  }
+  return sel
+}
+
+// Returns the architecture of the node a label has to build on.
+def archOf(String label) {
+  return label.contains('_arm_64') ? 'arm64' : 'amd64'
```
craigcomstock, line 24: exotics in 3.24.x return solaris, hpux, aix?

Jenkinsfile
```diff
@@ -0,0 +1,252 @@
...
+// Get the revision to build at from <repo>_REV parameter.
+def revFor(String repo) {
+  def rev = params[repo.toUpperCase().replaceAll('-', '_') + '_REV']
+  return (rev && rev.trim()) ? normalizeRev(rev.trim()) : defaultRev(repo)
+}
+
+// Returns the refspec to fetch a repo with. Branch heads are always fetched.
+// Pull requests live outside refs/heads. Building one means fetching its ref
+// too, or its commit is not in the clone.
+def refspecFor(String rev) {
```
craigcomstock, line 65: I wonder if there is a way we can share this boilerplate/functions around with other Jenkinsfiles in other repos. Maybe a follow up ticket. Of course the other strategy would be a template that then is rendered out to various repos.

Jenkinsfile
```diff
@@ -0,0 +1,252 @@
...
+  def heads = '+refs/heads/*:refs/remotes/origin/*'
+  if (!(rev ==~ /^(?:refs\/)?pull\/\d+\/(merge|head)$/)) { return heads }
```
craigcomstock, line 67: for things like this it would be cool to have unit tests ;) I know that is wacky in this context, but who knows? If we pushed these bits to a groovy script we might be able to unit test them with groovy.

Jenkinsfile
```diff
@@ -0,0 +1,252 @@
...
+  withCredentials([sshUserPrivateKey(credentialsId: 'jenkins-build-artifacts-cache',
+                                     keyFileVariable: 'JENKINS_SFTP_KEY_PATH')]) {
+    sh """set -eu
+          ./src/buildscripts/build-in-container.py ${opts} \\
+              --build-type '${buildType}' \\
+              --build-number '${params.PACKAGE_BUILD_NUMBER}' \\
+              --source-dir "\$WORKSPACE/src" \\
+              --output-dir "\$WORKSPACE/output" \\
+              --cache-dir '${cacheDir}' \\
+              --sftp-key "\$JENKINS_SFTP_KEY_PATH"
```
craigcomstock, line 115: Does the build also need the SFTP_PUBKEY for known_hosts?

craigcomstock (approved):
looks cool! thanks!

Jenkinsfile
larsewi: Yeah, these container builds wont work on exotics

Jenkinsfile
larsewi: Please elaborate :)

Jenkinsfile
larsewi: ENT-14457

PR #2446, opened 2026-08-20
Added workflow to mirror the ci directory to LTS branches

larsewi:
Copies `ci/` from `master` to `3.27.x` and `3.24.x` and opens a pull request for each. Runs on pushes to `master` that touch `ci/**`.

### DIFF
--- a/.github/workflows/mirror-ci.yml
+++ b/.github/workflows/mirror-ci.yml
@@ -1,0 +1,53 @@
+name: Mirror ci directory
+
+on:
+  push:
+    branches:
+      - master
+    paths:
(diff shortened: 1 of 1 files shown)


larsewi:
@craigcomstock does it make sense to mirror these files?

craigcomstock:
> @craigcomstock does it make sense to mirror these files?

In most cases yes... we do need to adjust some things based on which LTS we are operating in so if we do that in the scripts we can just mirror straight. Things I can think of are whether to vendor openssl, that sort of thing.

craigcomstock (approved):

PR #2450, opened 2026-08-21
CFE-3882: Added date binary inside cfengine

victormlg:


cf-bottom:

nickanderson (approved):
Seems reasonable, though I still wish we didn't have to exec out

craigcomstock (approved):

deps-packaging/coreutils/mingw/compat-include/pwd.h
```diff
@@ -0,0 +1,30 @@
+/* Minimal stand-in for <pwd.h> on native Windows (mingw-w64), which has no
+ * POSIX user database and does not ship this header at all.
+ *
+ * Several coreutils lib source files (idcache.c, userspec.c, ...)
+ * unconditionally include <pwd.h>, even though the functionality it
```
larsewi: The proper fix would be to patch the source to not include these unconditionally. If the assertion is wrong, that would surface this during compile time, instead of hiding it.

larsewi (changes requested):

deps-packaging/coreutils/0001-Guard-pwd.h-and-grp.h-includes-in-idcache.c-and-use.patch
```diff
@@ -0,0 +1,116 @@
...
+@@ -97,6 +101,14 @@ is_number (const char *str)
+ }
+ #endif
+
++/* Resolving a user/group spec requires looking it up via the system's
++   user and group databases, so the code below needs <pwd.h> and
++   <grp.h>.  On platforms that lack these headers (e.g. native Windows),
++   omit it entirely rather than fake up the lookups: if some program
++   actually needs it, it will fail to link, which is preferable to
++   silently linking in a passwd/group lookup that can never work.  */
```
larsewi, line 100: People reading this in the future will have no context of the fake look-ups that used to be there.

deps-packaging/coreutils/0001-Guard-pwd.h-and-grp.h-includes-in-idcache.c-and-use.patch
```diff
@@ -0,0 +1,116 @@
...
+references any symbol from these two files.
+
+Guard the includes, and the function bodies that depend on them, with
+the HAVE_PWD_H/HAVE_GRP_H macros that configure already defines via
+AC_CHECK_HEADERS. This way the two files compile down to empty
+translation units on platforms lacking these headers. If some other
+coreutils program actually needs the passwd/group lookups these files
+provide, linking that program will now fail with an undefined
+reference, surfacing the problem at build time instead of silently
+linking in stub declarations that could never work.
```
larsewi, line 23: This you could probably explain with a few sentences.

larsewi (approved):

PR #2461, opened 2026-08-26
ENT-14451: Suppressed errexit when sourcing /etc/profile in ci/fix-buildhost.sh

aleksandrychev:
On suse /etc/profile calls tty, which exits non-zero in a non-interactive CI shell.

Ticket: ENT-14451


craigcomstock (approved):
odd, I thought we had already adjusted for suse. This looks OK though.

aleksandrychev:
> odd, I thought we had already adjusted for suse. This looks OK though.

yeah, I saw you have added condition below, but this is still sits unconditionally

PR #2468, opened 2026-08-28
Jenkinsfile: Now checks out ref names instead of pinned commits

larsewi:
GitHub replaces refs/pull/N/merge every time the pull request is pushed to.
A build that had already pinned the old merge commit could then no longer
check it out.

The other Jenkins jobs pass the ref name to git and let each node resolve
it, so do that here too.


cf-bottom:

craigcomstock (approved):

PR #2500, opened 2026-09-16
ent-9880: package cfengine enterprise client for macos

craigcomstock:
todo
- [ ] pull package out of builder and make an artifact on github summary


craigcomstock:
I will add a github workflow and in that use a script, maybe in ci directory for what steps to follow to make the build.

